PR #18, opened 2026-06-03
WIP: Stripe billing integration (Checkout + Portal + metered usage queue)

ewyct:
https://claude.ai/code/session_01X4572vhHQas5GtJ6RmmSFz


ewyct:
please check:
* Stripe Tax if it's as planned
* Refunds are manual now, we can add a 1 day "trial" as a technicality if we want misclick refunds auto
* Products in Stripe Test mode are different from Prod (needed to change to tax exclusive, EA pricing)

package.json
```diff
@@ -4,6 +4,9 @@
   "private": true,
   "license": "FSL-1.1-ALv2",
   "pnpm": {
+    "overrides": {
+      "kysely": "^0.28.17"
+    },
```
mtolmacs: Please rebase or merge main and remove this because main now has the correct version override.

Also overrides should always go to pnpm-workspace.yaml, having pnpm overrides in package.json is deprecated.

.env.example
```diff
@@ -48,4 +48,79 @@ NEXT_PUBLIC_UMAMI_WEBSITE_ID=
...
+#     Checkout page renders Stripe-orchestrated payment methods + 3DS.
+#
+STRIPE_SECRET_KEY=
+STRIPE_WEBHOOK_SECRET=
+
+# Set to "true" to collect tax on Checkout via Stripe Tax
+# (automatic_tax). Requires Stripe Tax to be activated on the account
+# AND at least one tax registration (Dashboard → Settings → Tax) —
+# checkout errors if enabled without that setup, hence opt-in.
+STRIPE_AUTOMATIC_TAX=
```
mtolmacs: Recommend removing all this automatic tax thing completely. We will not use it and it just adds cruft and dead code.

docs/gtm/saas-demo-log.md
mtolmacs, line 1: Please move this to a separate PR, this has no relevance to subscriptions

scripts/git-hooks/check-stripe-keys.sh
mtolmacs, line 1: Let's merge the Husky hooks and wire this check into that

src/lib/db/schema.ts
```diff
@@ -3617,3 +3640,79 @@ export const launchMonthlyWinners = pgTable('launch_monthly_winners', {
...
+  status: text('status').$type<SubscriptionStatus>().default('incomplete'),
+  periodStart: timestamp('periodStart'),
+  periodEnd: timestamp('periodEnd'),
+  trialStart: timestamp('trialStart'),
+  trialEnd: timestamp('trialEnd'),
+  cancelAtPeriodEnd: boolean('cancelAtPeriodEnd').default(false),
+  cancelAt: timestamp('cancelAt'),
+  canceledAt: timestamp('canceledAt'),
+  endedAt: timestamp('endedAt'),
+  seats: integer('seats'),
```
mtolmacs: Seats?!

src/server/actions/billing.test.ts
mtolmacs, line 1: Would love to have tests for these cases:
* Test trying to edit or create a subscription which is not owned by the user
* resumeTeamSubscription with active (non-cancelled) sub: What happens if you try to resume a sub that isn't cancelled?
* changeTeamPlan capability check: Same as above no capability enforcement test for changeTeamPlan
* changeTeamPlan with invalid plan tier: No test for passing a plan that doesn't exist (e.g. "enterprise") though this may be handled by the plugin
* Parallel/concurrent plan changes: No test for race conditions (two simultaneous changeTeamPlan calls). Should not allow parallel operations.
* Error propagation from plugin: callUpgradeSubscription / callCancelSubscription / callBillingPortal are always mocked with success. No test for when the plugin throws an error (network failure, Stripe API error, etc.)

mtolmacs (changes requested):
runTestsCore():

mtolmacs:
No referenceId ownership verification in server actions. The server actions pass session.team.id as the referenceId without an explicit assertion. If requireCapability ever changes to accept an external teamId, it would be an auth bypass. Consider adding:
```typescript
if (billing?.stripeSubscriptionId) {
  const sub = await queries.getSubscriptionByTeamId(session.team.id);
  if (sub?.referenceId !== session.team.id) throw new Error('...');
}
```

mtolmacs:
openCustomerPortal doesn't check for active subscription. It calls callBillingPortal without verifying the team has an active subscription. Stripe's portal may return an error or empty state for free-tier teams. Compare with cancelTeamSubscription which explicitly checks for stripeSubscriptionId.

mtolmacs:
handleSubscriptionUpdate silently ignores missing price IDs. When stripeSubscription.items.data[0]?.price.id is null, resolvePlanForPriceId returns null, and planId becomes undefined, so syncTeamPlanForBilling is never called (webhook-sync.ts:79). A plan change via Stripe portal could be silently ignored. Should log a warning or error.

mtolmacs:
Must verify what happens if trial ends and stripe transitions to paid subscription. Does this code properly records the change? Does better-auth/stripe hanle it?

mtolmacs:
Webhook log write failures don't block sync. The onEvent callback (auth.ts:322-331) catches errors from recordStripeWebhookReceipt and markStripeWebhookProcessed. This is correct for availability â the webhook sync (onSubscriptionComplete, etc.) runs before the log write in the plugin lifecycle, so the sync happens regardless. However, if the log write fails, admins lose their forensic record. Consider a separate retry queue for log writes rather than silently dropping them.

.env.example
ewyct: How are we charging 14EUR+tax then?

mtolmacs:
No migration for existing teams. If there are teams on the old teamBilling table, there's no migration to migrate stripeCustomerId, stripeSubscriptionId, etc. into the new subscriptions table. Teams that had subscriptions before this branch would lose their billing state.

.env.example
mtolmacs: We hopefully will not charge tax from customers at all. That's why we use Stripe Managed Payments exclusively. Stripe takes care of local customer-facing taxes. Stripe Tax is a separate product from Managed Payments.

mtolmacs:
Can you verify this? updateTeamStorageQuota and updateTeamRunQuota seem to have no authorization. These functions (storage.ts:38-43, 106-111) accept any teamId and quota they could be called by any authenticated user if exposed through an API. No requireCapability guard seems exist.

mtolmacs:
No rate limiting on billing actions. startCheckout, changeTeamPlan, and cancelTeamSubscription can be called repeatedly. While Stripe would eventually reject duplicate checkouts, the app should rate-limit these to prevent abuse (e.g., a user triggering 100 checkout sessions per minute).

mtolmacs:
No downgrade protection for past-due subscriptions: A team with past_due status can still call changeTeamPlan to upgrade, but Stripe would likely fail the proration charge. Should check subscription status before allowing plan changes.

src/lib/db/schema.ts
ewyct: The @better-auth/stripe plugin writes seats to the subscription table on every create/update; dropping the column would break the plugin's adapter.

mtolmacs:
Screenshot from 2026-06-07 19-56-24

Layout seems to be broken.

mtolmacs:
Seems like the organization plugin is missing? I tried the GitHub login/signup.
```
2026-06-07T18:26:20.250Z ERROR [Better Auth]: The field "activeOrganizationId" does not exist in the "session" Drizzle schema. Please update your drizzle schema or re-generate using "npx auth@latest generate".
# SERVER_ERROR:  [Error [BetterAuthError]: The field "activeOrganizationId" does not exist in the "session" Drizzle schema. Please update your drizzle schema or re-generate using "npx auth@latest generate".]
 GET /api/auth/callback/github?code=5edf519e3422d76dece7&state=SZII3FEpy_bzY9aa4uTGncBxOTsCMAIt 500 in 1606ms (compile: 113ms, proxy.ts: 3ms, render: 1490ms)
 ```

.env.example
ewyct: It's auto added by stripe currently.

ewyct:
> Screenshot from 2026-06-07 19-56-24
> Layout seems to be broken.

How did you get to this page? 
/billing shouldnt be accessible from ui and callbacks were removed to there.

mtolmacs:
> How did you get to this page? /billing shouldnt be accessible from ui and callbacks were removed to there.
Settings -> open full billing page small link

mtolmacs (changes requested):

mtolmacs (approved):

PR #19, opened 2026-06-04
Chore/package refactor

mtolmacs:
* Update potentially vulnerable packages
* Notable upgrades: next and better-auth

### DIFF
--- a/css.d.ts
+++ b/css.d.ts
@@ -1,0 +1,1 @@
+declare module "*.css";
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "stack:status": "bash scripts/k3d-status.sh"
   },
   "dependencies": {
-    "@anthropic-ai/claude-agent-sdk": "^0.2.98",
+    "@anthropic-ai/claude-agent-sdk": "^0.2.141",
     "@axe-core/playwright": "^4.11.0",
     "@dnd-kit/core": "^6.3.1",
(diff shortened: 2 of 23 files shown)


ewyct (dismissed):

mtolmacs:
The image before-after slider must be tested because ESLint complained about it and I replaced the  tags with next's <Image> tags @ewyct

ewyct:
On the share page breaks the before/after as expected.

## Error Type
Console Error

## Error Message
Encountered a script tag while rendering React component. Scripts inside React components are never executed when rendering on the client. Consider using template tag instead (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template).


    at script (<anonymous>:null:null)
    at PublicSharePage (src/app/(public)/r/[slug]/page.tsx:278:7)

## Code Frame
  276 |           nonce="" during hydration. The script has already loaded
  277 |           correctly under the original nonce; suppress the cosmetic mismatch. */}
> 278 |       <script src="/share-slider.js" defer nonce={nonce} suppressHydrationWarning />
      |       ^
  279 |       {videoSchema && (
  280 |         <script
  281 |           type="application/ld+json"

Next.js version: 16.2.6 (Turbopack)

ewyct:
on the build page the compare option doesnt render

ewyct:
Recording works without errors for me.

ewyct (approved):

PR #23, opened 2026-06-10
Claude/app security review zt5cqd

ewyct:
Fixed (15 of 17)
Critical

Host-process RCE — three paths (testSetupScript, QuickStart captureStorageState, play-agent testLoginScript) eval'd customer/AI code in the host process where DATABASE_URL/STRIPE_*/SYSTEM_EB_TOKEN live. All now execute in a disposable runner/EB pod via the existing executeSetupViaRunner path. Self-hosted installs with no pool keep an in-process fallback (acceptable — single-tenant).
High

AI keys no longer serialized to the browser (page uses the masking action; testAIConnection resolves masked keys server-side).
Team invites are now token-bound (acceptInvitation verifies the token was issued to the signed-in user's own email); sign-up never auto-joins by email match.
getStreamUrlForRunner now requires team ownership before returning the CDP stream URL + STREAM_AUTH_TOKEN; the runner→session lookup is no longer RPC-exposed.
"Analyze URL" and notification webhooks (custom/Slack/Discord + tester) now use the SSRF guard with per-redirect-hop revalidation and capped reflected bodies.
Medium

Three IDOR "use server" actions (computeChangeMap, autoApproveZeroDiffCases, triggerAIDiffAnalysis) relocated to plain lib modules — no longer RPC-invocable.
Hardcoded Discord webhook removed (env-only now).
Run-minute + project quotas enforced server-side behind ENFORCE_RUN_LIMITS.
EB-pod egress NetworkPolicy added to block cloud-metadata/link-local (DNS-rebinding defense).
Low — CSV formula injection neutralized; owner-invites restricted to owners; unused getAISettingsRaw removed; /api/stats header-only + timingSafeEqual.

Not done (per your instruction)
#13 token hashing — deferred.
#2 AI key masking — you said not needed, but it was already implemented and committed before that; I left it in place rather than revert a working fix.


ewyct:
broke down to chunks

PR #26, opened 2026-06-10
[Sec 03/17] Require team ownership for live CDP stream URL + token

ewyct:
**Severity: HIGH** — finding #4 of the security review.

`getStreamUrlForRunner` returned the live CDP stream URL **and the shared `STREAM_AUTH_TOKEN`** for any runner ID to any authenticated user — enough to watch (and via CDP, drive) another team's browser session.

It now requires team access and only serves sessions the caller's team owns (or shared system-pool EBs). The runner→session lookup is a private helper, no longer RPC-invocable.

---
**Stack 3/17 — base: `claude/sec-02-invite-token`.** Merge after Sec-02.

https://claude.ai/code/session_01QsD28d52asLNJX3YAsmjyV

---
_Generated by [Claude Code](https://claude.ai/code/session_01QsD28d52asLNJX3YAsmjyV)_


ewyct:

| 2 | 0 | 19 | 2 | 0 |

> **vs Main drift:** 12 tests have drifted from the main baseline

[View Build](http://localhost:3000/builds/95e825aa-59c8-490a-b7cd-830dfd25fe15)

---

src/server/actions/embedded-sessions.ts
```diff
@@ -380,9 +383,23 @@ export async function getStreamUrlForRunner(runnerId: string): Promise<{
...
+  // check, any authenticated user could pull another team's live CDP stream URL
+  // and the shared STREAM_AUTH_TOKEN.
+  if (session.teamId !== authed.team.id) {
+    const [runner] = await db
+      .select({ isSystem: runners.isSystem })
+      .from(runners)
+      .where(eq(runners.id, runnerId));
+    if (!runner?.isSystem) return null;
+  }
+
```
mtolmacs, line 402: Why would system EBs be accessible cross-team @ewyct ? If the user is running their test in a system EB, would it leak to other teams?

ewyct: a) there is bug and the activity view panel accidentally shares the agent MCP's stream to someone else
b) someone steals it somehow but no real value in that

due to a) i think this is worthwile.

mtolmacs: I'm not asking whether it's worth it, I just don't understand *why* would you want to skip team ownership test for system runners. Can't the system runners run tests for team users? Can't those tests be sensitive?

ewyct: The concept is that there are Team runners, potentially dedicated resources for an org, and there are system runners available for anyone (with a cap). Both have warm pool that exists even if no jobs are running. System runners can run tests for anyone, but team runners only do it for their team. At least this is the intended functionality on my end.

mtolmacs (approved):

PR #42, opened 2026-06-10
feat(security): encrypt oauth tokens and ai keys at rest

szuhaydv:


szuhaydv:
## Migration instructions (required before/after deploy)

  **1. Generate an encryption key and add it to your environment before deploying:**

      openssl rand -hex 32

  Add the output to `.env` (local) and your production environment:

      ENCRYPTION_KEY=<output>

  The app will throw on startup if this variable is missing or not exactly 64 hex characters.

  **2. Deploy the code first.** Existing plaintext rows continue to work — the app reads them transparently until backfilled.

  **3. Run the backfill against production after deploy:**

      # Dry run first — shows what will be encrypted without writing anything
      pnpm tsx --env-file=.env scripts/backfill-encrypt-tokens.ts --dry-run

      # Then apply
      pnpm tsx --env-file=.env scripts/backfill-encrypt-tokens.ts

  The backfill is idempotent — safe to re-run if interrupted.

  ---

  ## Algorithm: AES-256-GCM
  - Symmetric encryption using Node's built-in `crypto` — no new dependencies
  - 96-bit random IV per encryption, with an authentication tag (GCM provides tamper detection, not just confidentiality)
  - Key is a 32-byte (64 hex char) secret stored in the environment, never in the DB

  ## Sentinel prefix (`enc:v1:`)
  - Every encrypted value is prefixed with `enc:v1:` so the app can tell encrypted from plaintext
  - This is what makes zero-downtime migration possible — the decrypt function checks for the prefix and passes through plaintext values untouched if it's absent
  - The `v1` gives you a version escape hatch if you ever need to rotate the algorithm

  ## Where encryption/decryption happens
  - At the query layer boundary — encrypt on every write, decrypt on every read
  - Callers always see plaintext; the DB always sees ciphertext
  - No schema changes needed — columns stay as `text`

  ## Two write paths, two solutions
  - Our own query functions (`githubAccounts`, `gitlabAccounts`, `googleSheetsAccounts`, `aiSettings`) — wrap directly with `encryptField`/`decryptField`
  - `oauthAccounts` — owned by better-auth which bypasses our query functions entirely, so we used better-auth's `databaseHooks.account.before` to intercept and encrypt
  before the write. The `after` hook also needed updating to decrypt before passing the token to `syncGithubAccount` (which makes live GitHub API calls)

  ## Zero-downtime migration strategy
  - Deploy code first — existing plaintext rows keep working because `decryptField` passes them through
  - Run the backfill script after deploy — it reads every row, skips already-encrypted ones (`enc:v1:` check), encrypts the rest, and writes them back one at a time
  - No maintenance window, no user impact
  - Backfill is idempotent — safe to re-run

  ## What we deliberately skipped and why
  - `runners.tokenHash` — already a one-way hash, not a credential
  - `sessions.token`, `passwordResetTokens.token`, `emailVerificationTokens.token` — short-lived, one-time-use; encrypting them adds no security value
  - `oauthAccounts.password` — argon2 hash managed by better-auth; encrypting a hash would break auth verification

mtolmacs (changes requested):
Needs key rotation strategy with step-by-step documentation how to complete it. You can assume that the old key is available when the rotation happens.

mtolmacs (approved):

PR #44, opened 2026-06-10
Fix selector fallback learning loop: atomic stats, loser skipping, pa…

ewyct:
…rity

Selector-stats reliability:
- Add unique index on (test_id, hash, type, value); replace the racy SELECT-then-INSERT recorders with one batched atomic ON CONFLICT upsert (pre-aggregated per key), so concurrent runs can no longer duplicate rows or lose increments. recordSelectorSuccess/Failure become thin wrappers. scripts/dedupe-selector-stats.mjs merges legacy duplicates before db:push can create the index.
- Skip a missing responseTimeMs instead of folding 0 into the average.

Adaptive timeouts (selectorTimeoutFor):
- Known losers (3+ attempts, zero successes) now get min(default, 500ms) — previously they burned the full default because failures never record an avg, defeating the documented intent.
- The top-ranked candidate always gets the full default so one slow page load can't start a self-reinforcing demotion of the proven winner.

Selection quality:
- Record locate outcomes from waitFor alone; action failures (overlay, detach race) retry once on the matched element and no longer poison the selector ranking or silently act on a different element.
- relevantStats() falls back to per-(type,value) history across hashes, so a re-recorded/edited candidate array keeps its learned winner instead of starting cold.
- Anchor the 'undefined' interpolation-leak filter to token boundaries; values like #undefined-state-banner are no longer dropped (isUsableSelectorValue in @lastest/shared).

Executor parity (runner / EB test+setup / EB debug / host script-runner / recorded-code template):
- EB executors now resolve label / alt-text / title selectors (previously invalid-engine instant failures); runner now performs check/uncheck instead of recording a no-op success; EB coordinate-fill gains the 100ms focus settle; executors gain scrollIntoViewIfNeeded like the template.
- O(1) stat lookup via shared buildStatMap instead of per-candidate Array.find.

Note: run `node scripts/dedupe-selector-stats.mjs` then `pnpm db:push`; rebuild the EB image (pnpm stack:refresh) to pick up the executor changes.

https://claude.ai/code/session_01HVtRGEfdWtRpNJiFrqPjEF


ewyct:

| 2 | 0 | 16 | 5 | 0 |

> **vs Main drift:** 13 tests have drifted from the main baseline

[View Build](http://localhost:3000/builds/132f07d9-d455-4a5c-8902-28b05dc91e81)

---

mtolmacs (approved):

PR #49, opened 2026-06-12
Fix recording/debug session lifecycle and stream reliability issues

ewyct:
## Summary

This PR fixes multiple critical issues with recording and debug session management, stream reliability, and command deduplication that were causing sessions to restart unexpectedly, recordings to freeze, and resource leaks.

### Embedded Browser Session Management
- **Session ID tracking**: Added `currentRecordingSessionId` and `currentDebugSessionId` to prevent redispatched start commands from restarting live sessions
- **Command deduplication**: Implemented `isDuplicateCommand()` to drop redispatched commands after acking (mirrors server-side dedup in runner client)
- **Debug session cleanup**: When a new debug session arrives, properly tear down the previous one (stop executor, clear intervals) instead of leaking resources
- **Recording inactivity timeout**: Increased from 60s to 5 minutes and now counts live viewer input as activity (via `streamServer.onInputActivity`)
- **Watchdog improvements**: Clear previous interval before re-arming to prevent double-fire; route synthesized stop commands through `enqueueCommand()` for proper ordering

### Screencast Stream Reliability
- **Session tracking**: Store `currentPage` and validate session identity in frame callbacks to ignore frames from superseded sessions
- **Ack failure recovery**: Implement automatic recovery mechanism that restarts the screencast when consecutive ack failures indicate a broken CDP session, instead of silently freezing
- **Cleanup robustness**: Always tear down previous CDP session on `start()` (not just when `running` is true) to prevent session leaks and stale frame listeners

### Stream Server & Viewer
- **Input activity signal**: Added `onInputActivity` callback so recording inactivity watchdog doesn't kill sessions the user is actively viewing
- **Keepalive improvements**: 
  - Re-broadcast current status instead of hardcoded "idle" (was suppressing stall detection during recordings)
  - Fire every 2s instead of 5s (old cadence could deliver first keepalive ~8s after last frame, racing the viewer's 8s stall timeout)
- **File chooser handling**: Only honor explicit `fileChooserPending` values in status messages; keepalive messages omit the field and shouldn't clear the overlay
- **Viewport resize guard**: Only send initial viewport if explicitly provided; unconditional 1280×720 fallback was resizing live sessions on reconnect

### Command Ordering & Serialization
- **Command queue**: Added `commandChain` promise to serialize command execution so start/stop pairs can't interleave their teardown
- **Public enqueue method**: `enqueueCommand()` allows locally-synthesized commands (e.g., watchdog stop) to go through the same ordering chain

### Database & Event Ingestion
- **Recording event upserts**: Changed from `onConflictDoNothing()` to `onConflictDoUpdate()` on (sessionId, sequence) unique index
  - EB re-emits events when verification settles or thumbnails arrive; latest copy is canonical
  - Dedup within batch first to prevent Postgres multi-row upsert conflicts
- **Command status transitions**: Session-control commands (start_recording, start_debug, etc.) now complete on ack instead of sitting at 'claimed' forever
  - Prevents reaper from bouncing them back to 'pending' and redispatching into live sessions
  - Only result-bearing commands (run_test, run_setup) wait for result rows
- **Pool prewarming**: Respect interactive reservation cap to prevent builds from occupying recording/debug slots

### Utilities & Helpers
- **Stream token helper**: New `appendStreamToken()` utility correctly appends auth token with proper query string separator (fixes token being swallowed into proxy target param)
- **Job queue ordering**: Explicit `ORDER BY` when claiming oldest pending jobs (was returning arbitrary rows, allowing newer jobs to jump queue)

### Debug Executor
- **Resume queueing**: Buffer resume commands that arrive while a step is executing (previous step was still running) instead of silently dropping them
  - User actions (step_forward, run_to_end) during step execution are now queued and processed at the next checkpoint

## Notable Implementation Details

- Recording inactivity now counts live viewer input, fixing the "recording froze after 60s of reading" issue
-

https://claude.ai/code/session_01Ajb3LE2aoji1FJ3BFhmXwh

### DIFF
--- a/packages/embedded-browser/src/debug-executor.ts
+++ b/packages/embedded-browser/src/debug-executor.ts
@@ -96,6 +96,14 @@ class PauseController {
   private pendingResolve: (() => void) | null = null;
   private pendingReject: ((err: Error) => void) | null = null;
   private onPause: () => void;
+  // A resume that arrived while no checkpoint was waiting (the previous step
+  // was still executing). Without buffering, that user action (step_forward /
+  // run_to_end clicked mid-step) was silently dropped and the session looked
+  // unresponsive — the user had to click again.
+  private queuedResume: {
+    mode: "paused" | "running" | "run_to_step";
+    target?: number;
+  } | null = null;
 
   constructor(
     mode: "paused" | "running" | "run_to_step",
@@ -112,6 +120,18 @@ class PauseController {
     if (this.mode === "running") return;
     if (this.mode === "run_to_step" && stepIdx < this.target) return;
 
+    // Consume a resume that was issued while the previous step was running.
+    if (this.queuedResume) {
+      const q = this.queuedResume;
+      this.queuedResume = null;
+      this.mode = q.mode;
+      if (q.target !== undefined) this.target = q.target;
+      if (this.mode === "running") return;
+      if (this.mode === "run_to_step" && stepIdx < this.target) return;
+      // mode === "paused" means single-step: execute this step, pause at next.
+      if (this.mode === "paused") return;
+    }
+
     if (this.mode === "run_to_step") this.mode = "paused";
     this.onPause();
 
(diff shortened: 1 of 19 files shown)


ewyct:

| -4 | 0 | 19 | 5 | 5 |

[View Build](http://localhost:3000/builds/6fb8f0e3-63db-44e3-a33d-758703b8616f)

---

mtolmacs (approved):

PR #50, opened 2026-06-12
fix: fullscreen live view — black screen, lost tooltips, blank canvas after exit

ewyct:
Three fullscreen bugs in the headed live viewer:

- Entering fullscreen rendered everything near-black: the fullscreened wrapper used translucent bg-muted/50, which composited over the browser's black ::backdrop. Now opaque bg-background.

- Toggling fullscreen swapped between two separate JSX branches, so BrowserViewer unmounted/remounted (new WebSocket, blank canvas) — and since CDP only emits frames on repaint, the canvas stayed empty until the next step caused one. The section is now a single persistent tree that only swaps classNames, so the viewer instance (socket + canvas) survives the toggle. The stream-server additionally caches the last broadcast frame and replays it to (re)connecting clients, so any reconnect paints immediately instead of waiting for a repaint.

- Step-list hover panels never appeared in fullscreen: Radix portals content to document.body, which isn't rendered while another element is fullscreen. HoverCardContent now portals into document.fullscreenElement when one is active.

https://claude.ai/code/session_01NSAQWHZzBZJJ1ewzfHwCLn

### DIFF
--- a/packages/embedded-browser/src/index.ts
+++ b/packages/embedded-browser/src/index.ts
@@ -492,7 +492,15 @@ async function startup(): Promise<void> {
         const isHeaded = !!payload.headed;
         if (activeTasks === 1) {
           capturedClient.setStatus("busy", payload.testId);
-          streamServer?.broadcastStatus("busy", payload.targetUrl);
+          // Headed playback gets a distinct "starting" status: the screencast
+          // is still bound to the idle page until onPageCreated re-attaches
+          // it to the test page, so viewers show a "Starting test…"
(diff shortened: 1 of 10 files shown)


ewyct:

| 3 | 0 | 15 | 3 | 4 |

> **vs Main drift:** 8 tests have drifted from the main baseline

[View Build](http://localhost:3000/builds/5dd08d70-6814-4a11-bb29-018bab868f7d)

---

mtolmacs (approved):

PR #55, opened 2026-06-16
Share video and dom improvements

ewyct:


ewyct:

| -1 | 0 | 29 | 0 | 3 |

[View Build](http://localhost:3000/builds/c34c2bab-1280-4d8b-a3c4-8aa451188484)

---

src/lib/share/video-mp4.ts
mtolmacs, line 1: Really, really not optimal. First, why do you need to transcode in the first place? You seem to use it for web video playback, [all modern browsers support webm out of the box](https://caniuse.com/webm).

But if you really need to trancode, ideally transcoding should happen in a separate service, not spawned by the next process. This is unstable and could make the node process really resource constrained. 

Even if you can't pull off a separate service, I really recommend using a more stable API to work with ffmpeg, like https://github.com/seydx/node-av (npmjs node-av).

.lintstagedrc.json
```diff
@@ -1,7 +1,7 @@
 {
-  "{*.ts,*.tsx}": ["pnpm prettier --list-different", "pnpm eslint"],
+  "{*.ts,*.tsx}": ["pnpm prettier --write", "pnpm eslint"],
   "{*.js,*.jsx,*.json,*.css,*.scss,*.mjs}": [
-    "pnpm prettier --list-different",
+    "pnpm prettier --write",
     "pnpm eslint"
```
mtolmacs, line 5: Don't do this, it makes you transform your files during commit, which is really not ideal. Just run it the write before you commit.

src/app/(public)/share/[slug]/download/route.ts
```diff
@@ -0,0 +1,97 @@
...
+    .replace(/^-+|-+$/g, "")
+    .slice(0, 60);
+  return base || "recording";
+}
+
+function normalize(p: string): string {
+  return "/" + p.replace(/^\/+/, "");
+}
+
+export async function GET(
```
Comment (mtolmacs) [inline src/app/(public)/share/[slug]/download/route.ts:34]: Wouldn't it need to have logged in session check? Or all sharing is public if you have the url?

mtolmacs (changes requested):
Mainly see my transcoding comment.

PR #61, opened 2026-06-17
QuickStart: surface the published /r/ share (streamline)

ewyct:
## QuickStart — make it a first-class, self-serve feature

QuickStart runs a 9-step agent pipeline against a repo's base URL and publishes a public `/r/<slug>` founder report. This PR makes that value visible, lowers the barrier to running it, gives the report a presentable graded framing, and promotes the feature out of Early Adopter.

### Done in this branch

**1. Surface the published share** — the client dropped the serialized `shareUrl`/`shareSlug` on the floor. Now a prominent **"Founder share ready"** block appears on completion with `/r/<slug>` + **Open report** + **Copy link**. Completed the step-label map.

**2. Inline base-URL entry** — the `no_base_url` empty state no longer punts you to the sidebar. It takes the repo's `defaultBranch` and writes straight to the gate's key (`saveBranchBaseUrl`), then `router.refresh()`es so the gate re-evaluates in place.

**3. Presentable graded report card on the share** — the share showed raw a11y/perf *numbers*; now graded letter-grade tiles via `scoreGrade()` + `GradeCard` (reuses `StatCard`):
- **Accessible** — `build.a11yScore` (WCAG 2.2)
- **Design** — `build.designSystemScore` (when present)
- **Fast** — absolute Core Web Vitals from `test_results.webVitals`, scored against Google's standard good/NI/poor thresholds (LCP 2500/4000ms, INP 200/500ms, CLS 0.1/0.25). Chosen over the perf *diff* layer, which is regression-only and empty on baseline-less first runs. Renders "—" when no vitals were captured.
- Grids size themselves to the tiles present (`gridColsClass`). Both build-wide and test-scoped layouts.
- Only new data plumbing: projecting `webVitals` into `ShareTestResult`. No schema/DB change.

**4. Dashboard: hide the Play Agent** (Auto Setup Agent) timeline.

**5. Promote QuickStart out of Early Adopter** — dropped the `earlyAdopterMode` gate in `isQuickstartEnabled` and the dashboard mount condition; removed the "early adopter" badges. QuickStart now only requires a non-local base URL (and not `banAiMode`).

All changes typecheck (`tsc --noEmit`, Node 20) + lint clean. `gating.test.ts` (covers `pickRepoBaseUrl`) still passes.

### Heads-up
Promoting out of EA exposes QuickStart to every team — each run consumes an EB pod + AI. Worth a glance at provisioning/quota before this lands.

### Why (competitive context)
Closest competitor `prufa.dev` productized our manual `/gtm-lastest-saas-demo` motion into a self-serve, no-login audit returning a shareable **graded** report card (Works/Fast/Accessible/Compliant…). This PR gives QuickStart the same one-tap-share + graded framing, built on our existing a11y/design/perf verify layers.

### Possible follow-ups
- [ ] Build-level `perf_score` column (today computed on-read from per-result vitals).
- [ ] SSE instead of 2.5s polling for the live timeline.
- [ ] AEO check (robots.txt AI-crawler block) in the scout.
- [ ] Remove now-dead `not_early_adopter` reason from the gate union/hint (left as a harmless superset).

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -97,8 +97,12 @@ const nextConfig: NextConfig = {
           source: "/_umami/recorder.js",
           destination: `${umamiUrl}/recorder.js`,
         },
-        { source: "/_umami/api/send", destination: `${umamiUrl}/api/send` },
-        { source: "/_umami/api/record", destination: `${umamiUrl}/api/record` },
+        // Ingest beacons go through a resilient route handler that ACKs the
+        // browser immediately and forwards to umami in the background, so a slow
+        // umami can never stall navigation (e.g. submitting the login form).
+        // The handler reads UMAMI_INTERNAL_URL itself at runtime.
+        { source: "/_umami/api/send", destination: "/api/umami/send" },
+        { source: "/_umami/api/record", destination: "/api/umami/record" },
       );
     }
 
--- a/packages/mcp-server/server.json
+++ b/packages/mcp-server/server.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-09-29/server.schema.json",
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-12-11/server.schema.json",
   "name": "io.github.las-team/lastest",
   "description": "Run visual regression tests, review diffs, and manage baselines on a Lastest instance.",
   "repository": {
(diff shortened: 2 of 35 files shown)


ewyct:

| -4 | 0 | 29 | 0 | 7 |

[View Build](http://localhost:3000/builds/8ef71826-2d75-4528-b936-6d2e5fd645d6)

---

mtolmacs (commented):
Not directly related, but wouldn't the user provided credentials for their own app need encryption? I can't find it now whether it is encrypted or not. Otherwise looks good.

ewyct:
> Not directly related, but wouldn't the user provided credentials for their own app need encryption? I can't find it now whether it is encrypted or not. Otherwise looks good.

No, but it was long due to be added (only user pass was encrypted), added a anew commit for it.

mtolmacs (approved):

PR #78, opened 2026-07-08
fix(oauth): register playground-www client so playground login works

ewyct:
## Problem

`/playground` leaderboard sign-in is fully broken in production. Clicking **Sign in with Google** redirects the browser to:

```
https://app.lastest.cloud/oauth/authorize?client_id=playground-www&...&provider=google&screen_hint=signup
```

…and the server responds `HTTP 400 {"error":"invalid_client"}` — **before the flow ever reaches Google**. The `provider=google` / `screen_hint=signup` params are red herrings; the request dies at the very first check (`isValidClientId`) in the authorize route.

**Root cause:** only `launch-www` was a registered OAuth client on `main`. `playground-www` was never registered in production, so the authorize endpoint rejects it exactly like a nonexistent client. Verified live: `launch-www` → `307` (works), `playground-www` → `400 invalid_client` (same as a bogus client id).

## Fix

- Register `playground-www` with its own `playground:score` scope via a `clientId → scope` map.
- Grant **per-client** scope in the authorize route instead of the hardcoded `LAUNCH_SCOPE`. This matters: a playground token must carry `playground:score`, not `launch:vote launch:submit` — otherwise the `/api/v1/playground/*` scope checks would reject it.
- Add `playground.lastest.cloud` to the redirect-origin allowlist (the actual redirect to `https://lastest.cloud/...` was already allowed via the apex origin).

Matches the frontend contract in `PLAYGROUND_BACKEND_SPEC.md` §3, which the already-shipped `lastest-www` playground frontend is wired to.

## Verification

- `vitest run src/lib/launch/oauth-config.test.ts` → 6 passed (adds coverage for `playground-www` validity + per-client scope mapping)
- `tsc --noEmit` → clean
- `eslint` (changed files) → clean

Once merged + deployed, `client_id=playground-www` will authorize and the playground leaderboard sign-in lights up.

### DIFF
--- a/src/app/api/v1/launch/[...path]/route.ts
+++ b/src/app/api/v1/launch/[...path]/route.ts
@@ -23,7 +23,6 @@
 
 import { NextRequest, NextResponse } from "next/server";
 import * as queries from "@/lib/db/queries";
-import { getCurrentSession } from "@/lib/auth";
 import { extractSourceIp } from "@/lib/url-diff/ssrf";
 import {
   assertCanVote,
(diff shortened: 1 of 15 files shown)


ewyct:

| 13 | 0 | 0 | 0 | 24 |

[View Build](http://localhost:3000/builds/3b596ad6-7d97-462a-8e30-829d40fbe83c)

---

mtolmacs (approved):

PR #89, opened 2026-07-20
Replace activity-feed WebSocket with existing SSE endpoint

ewyct:
## Summary

The activity-feed had two parallel broadcast implementations of the same `subscribeToActivityFeed()` source: a standalone WS server on port 9400 (server→client only — it never registered an `on("message")` handler) and the SSE endpoint at `/api/activity-feed` with identical repo/source filters and both cookie + bearer auth.

This deletes the WS layer and points both clients at the SSE endpoint:

- **Deleted** `src/lib/ws/activity-feed-server.ts` and its boot call in `instrumentation.ts`
- **Deleted** the `/api/activity-feed/ws` branch in `ws-proxy-preload.js`, plus `ACTIVITY_FEED_WS_PORT` and the `forwardCookie` plumbing only it used — the proxy now handles exactly one route (the grant-authenticated EB stream)
- **Removed** `"ws"` from `serverExternalPackages` (no app code imports it anymore); the Dockerfile `ws` copy stays for the bundled embedded-browser dist
- **Switched** `activity-feed-provider-client.tsx` and `use-activity-feed.ts` to `EventSource("/api/activity-feed")` with the same `repo`/`source` params. The hand-rolled retry/backoff loops are gone — EventSource auto-reconnects, including across the endpoint's 90s Cloudflare lifetime-cap close. Event parsing, dedupe, session-count, and toast logic unchanged.

## Rollout safety

The SSE endpoint already exists on currently deployed pods with identical filters and auth, so new client JS works against an old pod. An old cached tab hitting a new pod fails its `/ws` upgrade quietly and recovers on refresh. No deploy-ordering constraint.

## Verification

- `tsc --noEmit` and eslint clean
- Edited preload exercised directly with node: valid grants still route, bad grants 403, `/api/activity-feed/ws` now falls through as unhandled
- Note: vitest currently fails to resolve `@/` imports on this branch even on an unmodified tree (pre-existing, unrelated)


ewyct:

| 0 | 0 | 0 | 0 | 38 |

[View Build](http://localhost:3000/builds/6ace8e37-9381-400c-afa1-e3975c6ae919)

---

mtolmacs (approved):

PR #90, opened 2026-07-21
feat/websocket reverse proxy 8fy493

ewyct:
Yes — but a tiny one we own, not an external one like Traefik or nginx.

[scripts/front-proxy.js](https://github.com/las-team/lastest/blob/claude/websocket-reverse-proxy-8fy493/scripts/front-proxy.js) is a ~450-line dependency-free Node reverse proxy that now owns :3000 in every deployment. It routes exactly one thing specially — /api/embedded/stream/ws upgrades go straight to the EB pod named in the signed grant — and everything else (all HTTP, plus other websockets like dev HMR) is forwarded to Next, which now listens privately on 127.0.0.1:3001.

Why not a "real" reverse proxy? Two reasons, both from your constraints:

The routing decision is dynamic. The WS upstream is a per-session EB pod IP carried inside the HMAC-signed grant. Traefik/nginx/Envoy route on static rules (path → service); none of them can express "connect to whatever pod this signed token says." You'd need a custom plugin or a lookup sidecar — more moving parts than the 450 lines.
Your deployment spread. Zima self-hosters run bare docker-compose with no ingress, and on Olares the Envoy sidecar belongs to the platform, not you. An in-process proxy ships inside the image, so all targets work with zero manifest changes — same image, same :3000, same env.
The thing that actually got removed is the hack: the old preload patched http.Server.prototype.listen and shimmed socket methods to fight Next's own upgrade handler on a shared server. Now there's no shared server — the proxy owns the socket, Next never sees the upgrade, so all that defensive code is gone.

And it composes with a real edge proxy later: if you put Traefik in front on Kubernetes and route the WS path at the edge someday, the front proxy just never receives those upgrades — nothing to change or configure.


ewyct:
Benefits confirmed by testing:

Process provisioner is a big dev win: zero-infra pnpm dev (no k3d cluster, no 930MB image rebuild per EB change), per-session bootstrap tokens instead of a fleet secret, and EB children get no DATABASE_URL/ENCRYPTION_KEY
Kubernetes mode verified end-to-end locally (k3d): pool service in mode: kubernetes, EB provisioned, registered via bootstrap token, run executed. One migration note: clusters created by the pre-branch k3d-up.sh publish host :3001, which collides with the front-proxy's internal Next port — recreate the cluster or set UPSTREAM_PORT
What prod (Olares) actually runs and what we learned testing there:

Prod still runs the pre-branch image; nothing from this PR is deployed yet
Prod's two-Deployment split (lastest-dev + envoy-less lastest-internal-dev) exists because the Olares envoy sidecar is injected into whatever pods the entrance Service selects, and its ext_authz can't be path-exempted — verified in app-service source and by live experiment
We validated a replacement topology on a test app (prod image, chart-only): tiny envoyed nginx as the entrance, single envoy-less app pod behind it. Full test run + video recording passed. This means:
The PR's deferred gateway consolidation (EBs → pool-service only) is not needed for Olares — EBs reach the app directly
With one app pod, only one pool-service singleton starts — no EB_POOL_SERVICE_DISABLED coordination needed
Deploy prerequisites for this PR on Olares: ship app + EB images together (dynamic EBs only get EB_BOOTSTRAP_TOKEN now; old EB images can't register), and drop the chart's ws-proxy-preload.js args override since the image CMD becomes the front-proxy

mtolmacs (approved):

PR #97, opened 2026-07-31
Feat/pharma readyness

ewyct:


ewyct:

| -3 | 0 | 32 | 0 | 10 |

[View Build](http://localhost:3000/builds/243fd514-16f3-46f4-a28c-0413d01fb97e)

---

mtolmacs:
@ewyct 

* A cross-tenant IDOR: the coverage cell/dimension mutation actions authorize one repo but update rows by raw id with no repo scoping and no RLS.
* The matrix-execution core has several load-bearing gaps: expansion is silently discarded when the test count doesn't change (found independently by 4 finders — tests then run against the wrong data row with no coverage attribution), the visual policy flag (capturesVisual) and cell-scoped baselines (dataCell) are both computed/declared but never wired into dispatch or diffing, the API test path ignores matrices entirely, and build progress denominators are pre-expansion so progress can read "12 of 3".
* Two bugs make the coverage stopping rule permanently wrong: a lifetime covered-cell count is passed as the session run budget (any repo with ≥500 covered cells trips budget_exhausted forever), and build-written snapshots keep resetting the staleness clock so full re-profiling is starved on any active repo. Compounding it, the QA planner can't actually honor a stop — Math.max(1, …) plus empty-plan rejection guarantees at least one planned item.
* Two model bugs verified by execution: parseCoordsKey's unescape ordering corrupts backslash values (latent — no production callers yet), and the churn matcher misses standalone mentions when the first occurrence is embedded in a longer identifier.

mtolmacs:
* `expandMatrix` in the build path [matrix.ts:231]: `runBuildAsync` runs in the web process, and expansion synchronously materializes the full cartesian product before any cap: two 5,000-row sources → 25M objects → multi-second freeze or OOM of the whole app. This is the sharpest risk on the branch.
* The Coverage page itself [page.tsx:86]:  force-dynamic, double full-table fetch plus an uncached 20k-row trajectory-jsonb pull per view; also, syncCoverageAction hangs the clicking user for an entire sync.

[Here're all the findings](https://claude.ai/code/artifact/a5fcdc23-ff0f-40b8-abe0-4eeeec2e0766)

mtolmacs:
Not all highlighted issues were addressed fully.

* dataCell baselines are only half-wired (medium). Reads and the manual approve path are cell-aware, but the build-side auto-approve still deactivates every cell's baseline and creates a shared one — builds.ts:2515-2530 and builds.ts:2663-2678 pass no dataCell to deactivateBaselines / createBaseline. Under visual: "all" with forceAutoApprove or autoApproveDefaultBranch, the last cell to finish becomes the shared baseline and every other cell diffs "changed" on the next build. Pass dataCell in both places.

* Related design tension: core.ts:66-71 now makes the representative's approved baseline cell-scoped, while the getBranchBaseline doc (visual-diffs.ts:327-330) says the other runs "compare against the one baseline the representative established" — they can't, a cell-scoped baseline is invisible to other cells. That's harmless while non-representatives skip visual, but if the representative shifts (a new row reorders the pairwise pick) the new one has no baseline and reports "new test" forever. Under representative policy the approve should probably write the shared (NULL-cell) baseline.

* expandMatrix bound (medium, design). The 5,000 cap is real and the 25M-product test finishes in <5s. But truncation is positional (matrix.ts:262-275): for two 5,000-row sources the materialized set is axis0[0] × axis1[0..5000), so pairwise covering never varies the first source. The explanation is honest about it ("narrow the row filter"), but stride/interleaved sampling would give pairwise something to work with.

* Staleness fix unmasks the scheduler's inline batch (medium). Filtering to source: "sync" is right, but the audit warned to bound the tick in the same change and the commit doesn't touch scheduler.ts:181-212: every repo with enabled dimensions now gets a full syncCoverage (synchronous CSV parse, per-cell UPDATE loop) every 360 minutes, sequentially inside the web process. The new coverage_sync background job would be the natural target for the scheduler to enqueue into instead of running inline. E1 (full report computed and discarded every minute on the fresh path) is also still open.

* runsSoFar (low). Now defaults to 0 and no caller passes it (coverage-reads.ts:26 accepts only environmentKey), so budget_exhausted is unreachable and the maxRuns policy knob does nothing. Since planning runs once per session, 0 is defensible at plan time — but either wire the session count or document that maxRuns is inert.

* Background sync (low). revalidatePath inside the detached IIFE (coverage.ts:89) fires after the response is sent, so it's a no-op (the client's router.refresh() covers it). There's no per-repo in-flight dedup, so a reload-and-click or a coinciding scheduler tick can run two syncs concurrently through reconcile/prune — pre-existing, but easier to hit now. syncCoverageAction stays exported as a live RPC with no UI caller.

* API variables (low). The api path discards nextCursors/nextAiLastValues, so increment/random-cursor variables on api tests never advance between runs (browser path persists them at executor.ts:1079-1100).

mtolmacs:
Beyond the specifics above, I'm concerned that you run multi-seconds workflow items in the web process, adding significant load to the node or even DoS-ing it. Granted it needs a sessioned user, but still it can happen by accident:

# Concrete DoS vectors (authenticated user, own team)
startCoverageSyncAction in a loop — the client busy flag is the only throttle. Each call spawns a full syncCoverage: synchronous parseCsvBuffer on the main thread (twice per sync), one awaited UPDATE per cell, a 50k-row backfill load. 20 concurrent syncs on one repo also race the reconcile/prune step against each other. This commit made this easier — the action now returns instantly, so nothing back-pressures the caller.
createAndRunBuild while the pool has a free slot — every call spawns a runBuildAsync; matrix expansion (now capped at 5,000 candidates per test, but still synchronous) and diffing run in-process per build. Once the pool fills, further calls become queued rows without limit.
AI actions (aiFixTest, generateTestsFromStories, aiScanRoutes, …) — held-open requests plus provider spend, no cap.
profileFromSutAction — drives outbound HTTP to a user-supplied host inline; unused by the UI but live as an RPC.
Because it's all one process, the blast radius is every tenant: 10 pool connections exhausted → page loads, SSE ticks, and EB stream auth all stall.

# What would close it
A per-team in-flight cap in createJob (count running jobs of that type for the team; reject or queue above N) plus a per-repo dedupe for coverage_sync — cheapest, covers builds and sync at once.
Wire the plugin_jobs worker loop (1–2 claimers, SKIP LOCKED) and route coverage sync and the scheduler's stale-repo re-syncs through it instead of inline.
A generic team-keyed rate limit on mutating server actions, reusing rate-limit/limiter.ts.
Move CSV parsing to a worker_threads worker so even a legitimate large sync can't freeze the loop.

mtolmacs:
getCoverageAttributionTimeline orders ran_at ASC then applies LIMIT 50000, so on a ledger past 50k rows the backfill sees the oldest 50k, not the newest — its "newest 200 builds" are then 200 old builds and recent builds are invisible to it. Independent of this fix, but it means on such ledgers the probe's window and the backfill's can diverge (the probe's doc comment scopes its guarantee accordingly). Flipping the timeline to a newest-50k window is the obvious fix, but it changes the reconstruction's cumulative-coverage semantics (points would undercount cells covered before the window), so it deserves a decision

mtolmacs (approved):

PR #109, opened 2026-08-07
poc(macos): Appium mac2driver spike

szuhaydv:
Appium + mac2driver works. Confirmed locally against WhatsApp: session attach, accessibility tree, element resolution, clicking, window screenshots. Full loop proven — resolve selector → click → assert.

Start with `packages/appium-mac2-poc/FINDINGS.md`. SETUP.md to reproduce.

Two gotchas:

* The obvious screenshot endpoint captures the whole desktop — 3.3MB with a live clock, so baselines diff forever. Use the window-scoped element screenshot: 67KB, deterministic.
* macOS modals overlay the view tree instead of replacing it, so "old text is gone" assertions give false failures. Assert on what the dialog added.

Check layers: 4 of 9 usable (visual + text cleanly, dom/a11y partial), 5 don't apply. Same capability-gating conclusion as the Maestro PoC.

No runner built — ticket says local confirmation only, and Maestro already proved the protocol is engine-agnostic.

Caveats:

* CI is the real blocker: needs a logged-in macOS GUI + Accessibility permission, can't containerise.
* Selector surface is thin — zero app-authored test IDs, ~59% coordinate-only. Risky for third-party apps if we scope recording.
* All findings from one app.

Deps pinned pre-2026-08-04 with `ignore-scripts` (ChainDrop); audited all 617 packages. `appium-mac2-driver@4.1.0` landed inside the window — pinned 4.0.5.


ewyct:

| -11 | 0 | 31 | 5 | 13 |

> **vs Main drift:** 33 tests have drifted from the main baseline

[View Build](http://localhost:3000/builds/7ee65b19-3647-41be-adc1-87ff9253725c)

---

PR #82, opened 2026-07-15
Add interactive test playback & storage state evidence layer

ewyct:
## Summary

Implements two major features for the Verify review screen and test playback:

1. **Interactive Test Playback (Spec 28)**: Adds step-annotated video timeline with click-to-seek, playback↔evidence sync bus, and speed presets up to 8x. Unifies all evidence (network, perf, steps, storage) onto a single video clock (ms since recording start).

2. **Storage State Evidence Layer (Spec 27)**: New "State" tab in Verify showing cookies and localStorage diffs vs baseline, with per-entry change tracking (added/removed/changed) and end-of-run snapshot browsing.

### Video Playback & Timeline
- **`src/components/video-player.tsx`**: Extended with `PlayerSegment` interface for per-step markers, segment strip rendering with action icons (navigate/click/fill/assert/wait/etc.), click-to-seek, and speed presets up to 8x
- **`src/components/playback-sync.tsx`** (new): Bidirectional sync bus for playback↔evidence coordination; panes subscribe to time updates and can drive the player via `seekTo()`
- **`src/components/replay-player.tsx`**: Broadcasts playback position as document event (`PLAYBACK_TIME_EVENT`) for server-rendered islands (chapter rail)
- **`src/lib/playback/step-timings.ts`** (new): Resolves per-step video-clock segments from persisted `StepTiming` rows, screenshot anchors, or even split fallback
- **`docs/specs/28-interactive-test-playback.md`** (new): Full spec for step-synced timeline, clock unification, and sync bus architecture

### Storage State Evidence
- **`src/lib/comparison/storage-state-diff.ts`** (new): Diff engine comparing cookies/localStorage by structural key (domain+path+name for cookies, origin+name for localStorage), hashed values to prevent leaking sensitive data
- **`src/app/(app)/verify/[buildId]/focus-view.tsx`**: 
  - Added "State" tab to `COMPARE_TABS`
  - `StatePane` component renders storage diffs and end-of-run snapshots with change badges (ADD/REMOVE/CHANGE)
  - `StepDetailHeader` component shows per-step metadata: action kind, duration, video time range, verdict chip, thumbnail
  - Keyboard shortcuts `[` / `]` to cycle evidence tabs
  - `classifyLayer` and `layerDelta` support storage state classification
- **`docs/specs/27-verify-review-evidence-ux.md`** (new): Full spec for Verify review upgrades including storage state, per-step metadata strip, and evidence sync

### Database & Schema
- **`src/lib/db/schema.ts`**: 
[...]

PR #81, opened 2026-07-15
Add App Map "Explore" feature with multi-EB swarm crawler

ewyct:
## Summary

Implements the App Map "Explore" feature, enabling users to launch a multi-EB crawler swarm to progressively discover and map an application's pages. Adds exploration UI (dialog, progress panel, flows/screens tabs), a shared frontier work queue for coordinating multiple explorers, and server-side orchestration to manage the crawl lifecycle.

### Core Exploration Infrastructure
- **`src/lib/qa-agent/explore.ts`** (new): `SharedFrontier` class — a deterministic, side-effect-free work queue for N explorers sharing one in-memory frontier. Handles deduplication by normalized href + canonical path, round-robin segment assignment, work-stealing, and budget enforcement (depth, page count, wall-clock deadline).
- **`src/lib/qa-agent/explore.test.ts`** (new): Unit tests for frontier logic (deduping, claiming, balancing).
- **`src/server/actions/qa-agent.ts`** (modified): Added `runQaDiscoverSwarm()` orchestrator that claims N EB pods, injects storage state/login, and loops frontier→goto→extract→add across all explorers in parallel. Includes auth-context extraction from free-text sign-in prose via AI.

### App Map UI & Views
- **`src/app/(app)/app-map/app-map-client.tsx`** (new, ~1200 lines): Main React Flow canvas with custom PageNode components, dagre layout, spanning-tree hierarchy, node selection, detail panel, and tab switching (Map/Screens/Flows).
- **`src/app/(app)/app-map/explore-dialog.tsx`** (new): Modal to configure exploration (depth, strategy, max time, parallel explorers, optional auth context).
- **`src/app/(app)/app-map/explore-progress-panel.tsx`** (new): Live progress overlay showing per-explorer status (claiming/exploring/blocked/done), page counts, current URLs, and EB screencasts.
- **`src/app/(app)/app-map/screens-gallery.tsx`** (new): Flat responsive gallery of all mapped pages with screenshots.
- **`src/app/(app)/app-map/flows-view.tsx`** (new): Tab showing named user journeys derived from test trajectories.
- **`src/app/(app)/app-map/flow-player.tsx`** (new): Step-by-step player for flows with prev/next navigation, progress dots, and action labels.
- **`src/app/(app)/app-map/node-detail-panel.tsx`** (new): Right-side detail panel for selected nodes (screenshot, coverage, flows, QA agent request button).
[...]

PR #72, opened 2026-07-04
Issue RCA

ewyct:
## Summary

Implements the "confirm-on-green" feature that automatically closes GitHub issues filed by Lastest when a re-run of the linked test comes back `safe_to_merge`. This completes the diff → issue → fix → re-run loop without manual intervention, fulfilling the product promise that "the issue closes itself once the re-run comes back safe_to_merge".

## Key Changes

**New confirm-on-green module** (`src/lib/verify/confirm-on-green.ts`):
- Runs fire-and-forget when a build finalizes as `safe_to_merge`
- Closes every open Lastest-filed issue (`auto`/`open` state, not `linked`) whose test ran fully green in this build
- Skips `improvement` tickets (they track intent gaps, not regressions)
- Flips DB state to `closed` before GitHub PATCH to prevent webhook loops
- Includes error recovery: restores link state if GitHub PATCH fails (revoked token, deleted issue, etc.)
- Posts a comment linking the green build before closing

**Issue filing & linking** (`src/server/actions/diffs.ts`):
- Tags visual diff issues with `verify` label so the issues webhook recognizes close events and auto-reruns
- Auto-assigns configured AI engineer (Settings → Notifications → Issue Tracker)
- Mirrors GitHub issue link onto `step_comparisons` table so verify board shows the ticket and webhook can find the case

**Webhook improvements** (`src/app/api/webhooks/github/route.ts`):
- Changed `getStepComparisonsByGithubIssue` to repo-scoped `getStepComparisonsByGithubIssueInRepo` (issue numbers repeat across repos)
- Filters out already-closed cases to prevent re-run loops when confirm-on-green closes the issue itself
- Improved error messaging to distinguish "already reconciled" from "no linked steps"

**Database queries** (`src/lib/db/queries/step-comparisons.ts`):
- Added `setStepComparisonIssue()` to link GitHub issue metadata to a step comparison
- Added `getOpenIssueStepsForTests()` for confirm-on-green to find candidates
- Added `getStepComparisonsByGithubIssueInRepo()` for repo-scoped webhook lookups

**Issue body composer** (`src/lib/integrations/github-issue-body.ts`):
- Added RCA (Root Cause Analysis) section rendering (`renderRcaDrill()`)
- Shows verdict (code vs test vs uncertain), ranked signals, changed files, and element-level causes with CSS deltas
- Included in both visual diff and verify case bodies (rides along even when visual layer is filtered)
[...]

PR #110, opened 2026-08-08
RFC: Core + plugin refactor plan

ewyct:
## What

An architecture RFC for splitting the codebase into a small, protected **core** and feature **plugins** (hard-separated workspace packages). Doc only — `docs/architecture/core-plugin-refactor.md`, no code changes.

Stacked at the end of the current stack: `main` → #81 → #82 → #72 → **this**.

## Why

At the current feature pace, quality can't hold without a core that isn't touched casually. The RFC treats two things as the actual deliverable:

- **R3** — a core change is *mechanically* a separate PR that gets reviewed properly.
- **R4** — plugin code may only reach the platform through core: no direct-to-EB, no direct `@lastest/db`, no direct AI provider SDK.

Package extraction is the means, not the goal.

## What the RFC found

- `src/lib/*` is already close to plugin-shaped — cross-feature imports are thin. What's missing is *enforcement* and an injected capability surface.
- The coupling that does exist is exactly the one called out: `chromium.connectOverCDP(cdpUrl)` at 10 call sites across 7 files (`qa-agent/{auth,crawl,explore}.ts`, `explorer/tester.ts`, `playwright/ranger.ts`, `actions/play-agent.ts`). Feature code receives a raw pod address and connects itself.
- The weight isn't in `src/lib/` — it's in `src/server/actions/` (37k LOC / 74 files), `src/app/`, and `src/components/` (64k LOC). A boundary that only covers `src/lib/` moves ~15% of a feature.
- `verify/check-modes.ts` already has a 9-value `CheckLayer` union with per-layer modes — a plugin registry with the extension point hard-coded. Cheapest possible proof the framework works.

## Shape of the plan

- **Phase 0 (~1 week, nothing moves):** CODEOWNERS on core, branch protection, a CI check that fails any PR touching both core and plugins, ESLint import-boundary rules in *warn* mode against today's layout, graph test. This delivers R3 on its own — if the rest is never executed, this phase still changes how changes get reviewed.
- **Phase 1:** three spikes. The blocking one: does a `"use server"` module inside a `transpilePackages` package produce a working server action? If not, there's a codegen fallback, and if that's fragile too the RFC says stop and re-scope to server-side-logic-only plugins.
- **Phase 2:** kernel + `core/browser`, pilot = `explorer` (~5.1k LOC, exercises EB + AI + jobs + own tables + API route + app route + components).
- **Phases 3–5:** check-layer plugins, then one PR per feature, `qa-agent` last.

## Things worth arguing with

- `withRawPage()` escape hatch in the browser capability — wrapping all of Playwright isn't realistic day one, so R4 isn't perfectly enforced at the start. The RFC argues for it (plugins never hold a pod address; claim/release/metering/teardown centralise; the hatch is countable and can be driven to zero) but it's the weakest point.
- The `src/lib/playwright` split (12.8k LOC) is the call most likely to be wrong — deliberately sequenced last.
- Cost is ~a quarter of part-time work, ~20 PRs. Phase 0 is structured so the main benefit doesn't wait for that.
- §11 lists 5 open decisions, incl. whether `billing` is core or plugin, and whether phase-0-only is an acceptable permanent end state.


ewyct:

| -10 | 0 | 28 | 8 | 12 |

> **vs Main drift:** 33 tests have drifted from the main baseline

[View Build](http://localhost:3000/builds/3cebc711-7313-4255-9272-7cbb74914f2c)

---

mtolmacs:
# Migrator image doesn't have access to schema
```
[migrate] Running drizzle-kit push...
Error: Cannot find module './schema/eb-protocol'
Require stack:
- /app/packages/db/src/schema.ts
[migrate] Done
```
drizzle-kit push crashed with MODULE_NOT_FOUND — packages/db/src/schema.ts imports ./schema/eb-protocol, which isn't in the lastest-migrate:pr-110 image. So the schema push never ran. The database has 2 tables out of production's 98, and both came from migrate.js's pre-create phase, not the push — data_sources_csv_sources is literally the rename step in the log, and every subsequent pre-create step logged skipped: relation does not exist.

Line 40 of Dockerfile.migrate:
```
COPY packages/db/src/schema.ts ./packages/db/src/schema.ts
```
It copies one file. PR 110 splits the schema into a directory:

packages/db/src/schema/ — 10 files including eb-protocol.ts — exists on the branch
the same directory does not exist on main (404), so it's new in this PR
schema.ts:25 now does export * from "./schema/eb-protocol"
So the image ships schema.ts with imports pointing at a schema/ directory that was never copied in. drizzle-kit push loads it, hits MODULE_NOT_FOUND, and dies.

The Dockerfile even explains the assumption it's now breaking, at lines 15–18: "only the schema file itself is copied; its cross-package imports vanish" — that reasoning holds for type-only @lastest/* imports, which esbuild erases. A export * from "./schema/eb-protocol" is a value re-export, so esbuild must resolve it, and there's nothing there to resolve.

The fix is one line in the PR, changing line 40 to copy the directory:
```
COPY packages/db/src/schema.ts ./packages/db/src/schema.ts
COPY packages/db/src/schema/ ./packages/db/src/schema/
```

mtolmacs (approved):

PR #112, opened 2026-08-16
fix: stop mounting service account tokens in EB pods

mtolmacs:
EB pods run untrusted user/AI-authored Playwright code but the provisioned Job spec set no serviceAccountName or automountServiceAccountToken, so every pod ran with the namespace default ServiceAccount token mounted. On clusters where the default SA holds any RBAC, a compromised EB (or a Chromium sandbox escape) could reach the Kubernetes API.

EBs need no cluster API access — they only call back to the app over LASTEST_URL — so set automountServiceAccountToken: false in jobSpec() and in the reference k8s/embedded-browser-job.yaml template. Adds a regression test on the Job spec.


ewyct:

| -47 | 0 | 48 | 5 | 32 |

[View Build](http://localhost:3000/builds/38e78961-bad7-42bd-b2d2-24abd269d7e0)

---

PR #113, opened 2026-08-27
feat(agents): Agents console as the entry point for agent work

ewyct:
**Stack: 1 of 2.** Base is `feat/pharma-readyness`. The Explorer half is #114, stacked on this branch — core and plugin changes stay separate PRs per `core-scope.md` / RFC §7.2.

## What this adds

`/agents` — one roster of every agent working the selected repo, with what each is doing and what it is waiting on a human for. The sidebar's **QA Agent** entry becomes **Agents**; the QA agent page is now reached by drilling through a roster row and gains a breadcrumb back out.

## The idea worth reviewing

`paused` and `blocked on a human` are the same `agent_sessions.status` but opposite things to whoever reads the console:

- a **paused** run was stopped by a person and has released its embedded browser
- a **blocked** run is mid-flight, parked on a `waiting_user` step, and is **still holding** it

`deriveState()` splits them, and the capacity read-out counts held browsers — so an unanswered question shows up as pool capacity you cannot use. That is the argument for having a console at all, and it is the thing to push back on if you disagree.

Escalations merge two sources that are one thing to a human: a session on a `waiting_user` step, and a `qa_tasks` row the agent pushed back as `needs_input`.

## Notes for the reviewer

- **No new tables, no migration.** Everything is derived from `agent_sessions`, `qa_tasks` and the existing team run-usage counters.
- **Explorer is deliberately missing** from the roster here — `explorer_sessions` is the plugin's table and core cannot select from it. It keeps its own sidebar entry until the next PR in the stack.
- **The Pro gate moved onto `/agents`.** The console is the door to agent work, so it carries the entitlement check; `/qa-agent` keeps its own so direct navigation cannot bypass it. `QaAgentUpgradeGate` gained optional `title`/`icon`/`description` so the locked console heads with its own name — the run preview beneath is unchanged, since a QA run is what the plan unlocks.
- **`decryptAgentSessionRow` is now exported** from `queries/integrations.ts` so the new cross-kind selects apply the same decrypt-on-read as every other reader.
- **No live pool-service call** on page render. The console reports browsers held *by this repo's agents*; it does not fetch the fleet-wide cap, because that is a network hop to another process on every navigation.

## Not in scope

The design also had a per-agent policy strip (scope, time budget, guardrails). That needs a fleet-policy model that does not exist yet — building the strip now would mean showing invented numbers, so it is deferred rather than faked.

## Verification

- `pnpm exec vitest run` — 128 files, 1921 passing
- `pnpm exec tsc --noEmit` — clean
- `pnpm build` — compiled successfully, `/agents` present in the route table
- 12 new unit tests over the derivation rules (state split, browser accounting, narration, ordering, escalation merge)

### DIFF
--- a/plugins/qa-agent/src/data/tasks.ts
+++ b/plugins/qa-agent/src/data/tasks.ts
@@ -1,4 +1,4 @@
-import { and, asc, desc, eq, inArray } from "drizzle-orm";
+import { and, asc, count, desc, eq, inArray } from "drizzle-orm";
 
 import type { QaTaskStatus, QaTaskTestRef } from "../types";
 import { qaAgentTasks, type NewQaAgentTask, type QaAgentTask } from "../schema";
@@ -76,6 +76,44 @@ export async function getQaTasksByRepoRows(
       .limit(terminalLimit),
   ]);
   return [...open, ...terminal];
+}
+
+/**
+ * What the Agents console needs from the queue: the tasks pushed back for a
+ * human, plus how many are waiting to be picked up.
+ *
+ * Deliberately NOT `getQaTasksByRepoRows` — the console renders two numbers
+ * and a short list, and it is a `force-dynamic` page hit on every navigation.
+ * Pulling every task on the repo to filter two of them in JS is the shape
+ * flagged on #97; a filtered select plus a `count(*)` is the same answer for
+ * a fraction of the rows.
+ */
+export async function getQaConsoleQueueRows(
+  db: QaAgentDb,
+  repositoryId: string,
+): Promise<{ needsInput: QaAgentTask[]; queuedCount: number }> {
+  const [needsInput, queued] = await Promise.all([
+    db
+      .select()
+      .from(qaAgentTasks)
+      .where(
+        and(
+          eq(qaAgentTasks.repositoryId, repositoryId),
+          eq(qaAgentTasks.status, "needs_input"),
+        ),
+      )
+      .orderBy(desc(qaAgentTasks.updatedAt)),
+    db
+      .select({ n: count() })
+      .from(qaAgentTasks)
+      .where(
+        and(
+          eq(qaAgentTasks.repositoryId, repositoryId),
+          eq(qaAgentTasks.status, "queued"),
+        ),
+      ),
+  ]);
+  return { needsInput, queuedCount: Number(queued[0]?.n ?? 0) };
 }
 
 /** Oldest queued task for a repo — what the dispatcher picks up next. */
--- a/plugins/qa-agent/src/reads.ts
+++ b/plugins/qa-agent/src/reads.ts
@@ -1,5 +1,9 @@
 import { orm } from "./data/db";
-import { getQaTaskRow, getQaTasksByRepoRows } from "./data/tasks";
+import {
+  getQaConsoleQueueRows,
+  getQaTaskRow,
+  getQaTasksByRepoRows,
+} from "./data/tasks";
 import { getQaAgentTriggerRow } from "./data/triggers";
 import type { QaAgentTask, QaAgentTriggerRow } from "./schema";
 import { qaAgentWiring } from "./wiring";
@@ -22,6 +26,16 @@ export async function getQaTasksByRepo(
   opts: { terminalLimit?: number } = {},
 ): Promise<QaAgentTask[]> {
   return getQaTasksByRepoRows(orm(qaAgentWiring().data), repositoryId, opts);
+}
+
+/**
+ * The Agents console's slice of the queue — `needs_input` tasks and the queued
+ * count, without pulling the whole board. See `getQaConsoleQueueRows`.
+ */
+export async function getQaConsoleQueue(
+  repositoryId: string,
+): Promise<{ needsInput: QaAgentTask[]; queuedCount: number }> {
+  return getQaConsoleQueueRows(orm(qaAgentWiring().data), repositoryId);
 }
 
 /** One task by id (the v1 API's post-create echo). */
--- a/plugins/qa-agent/src/ui/qa-agent-upgrade-gate.tsx
+++ b/plugins/qa-agent/src/ui/qa-agent-upgrade-gate.tsx
@@ -1,3 +1,4 @@
+import type { ComponentType } from "react";
 import Link from "next/link";
 import {
   Badge,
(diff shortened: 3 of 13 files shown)


ewyct:

| -1 | 0 | 30 | 0 | 10 |

[View Build](http://localhost:3000/builds/97e90a38-53b4-4543-82ca-1b39b3969f97)

---

ewyct:
Reviewed against the finding classes from #97 (cross-tenant id handling, silently-discarded work, declared-but-never-wired fields, unbounded/uncached work on a `force-dynamic` page).

* **`listRecentSettledAgentSessions` sorts NULLs to the top** [`agents-fleet.ts`]: `orderBy(desc(completedAt), desc(createdAt))` — Postgres `DESC` is `NULLS FIRST`. Every settled row with no `completedAt` (cancelled runs, failures that never stamped it) outranks genuinely recent ones, so the "Recently settled" strip fills with the oldest junk. Needs `sql\`completed_at DESC NULLS LAST\``.
* Same function's docstring promises "the most recent settled session **per kind**"; the query is a flat `limit 8` with no partition. Doc and behaviour disagree about what the strip contains.
* **`holdsBrowser` is asserted, not measured** [`fleet.ts`]: derived purely from `state === "working" || "blocked"`. The console then renders "**N** browsers held" as a fact, and it is the one number on the page a user would act on (pool capacity). Nothing consults the EB pool. Either read the pool or label it as an estimate.
* **`getQaTasksByRepo(repo.id)` pulls every task on the repo** to compute two numbers — `queued` count and a `needs_input` filter — on a `force-dynamic` page, on every navigation. Same shape as the Coverage page fetch flagged on #97; a count + a filtered select would do.
* **The plan gate is conditional on `team` being present**: `if (team && !hasQaAgentAccess(...))`. A request that resolves no team skips the gate entirely and only fails to render because `getSelectedRepository(undefined, undefined)` returns null. The guard shouldn't depend on that coincidence.
* `escalationsFrom` sorts ascending on `since?.getTime() ?? 0`, so any escalation with a null timestamp pins permanently to the top of "Waiting on you".

mtolmacs (changes requested):
1. blocked state is unreachable — the console's core story silently collapses to "Paused"  at fleet.ts:141-144
deriveState returns "paused" before it ever looks for a waiting_user step. But every real human gate writes both together: QA plan review sets qa_plan_review = waiting_user then updateSession({ status: "paused" }) (actions.ts:2340-2344), and play-agent's setStepWaitingUser does the same (play-agent.ts:234-243). So for a real QA run at plan review, /agents shows "Paused", summary.blocked = 0, holdsBrowser = false, and the escalation panel says "Nothing is blocked" — the opposite of what /qa-agent shows for the same row. The unit tests pass only because the fixtures use status: "active" + waiting_user (fleet.test.ts:29-43), a combination no writer produces. Fix: check for a waiting_user step first (on active or paused), then fall through to paused/working.

2. e2e suite breaks: sidebar link selector no longer exists at sidebar.tsx:85
settings-ui.integration.test.ts:349 and :386 locate nav a[href='/qa-agent'] to assert the Pro lock badge; that anchor is gone, so both gating tests time out after 60s. They should target nav a[href='/agents'] (and ideally the /agents gate copy).

3. Escalation "waiting since" is the run start, not the gate time at fleet.ts:262
Session escalations use since: r.startedAt (= session.createdAt) while task escalations use updatedAt. A run that started at 09:00 and hit its gate at 12:00 sorts above (and shows "3h ago" next to) a task pushed back at 10:30 that has genuinely waited longer. AgentStepState.startedAt on the waiting_user step is the right source.

mtolmacs:
Other small things:

1. Run-minutes rendered as raw floats at agents-console-client.tsx:263
{runUsage.used} / {runUsage.quota}: run minutes are fractional (accumulated via COALESCE(...) + ${minutes}), so this will render e.g. 12.333333 / 1000 where the settings card renders 12.3 of 1,000 (run-usage-card-client.tsx:88-120). Reuse computeRunUsageProjection().usedPct and the existing formatting.

2. Roster shows a permanently idle "Ranger" row at agents-fleet.ts:25
FLEET_AGENT_KINDS includes "ranger", but Ranger sessions live in the plugin's own ranger_sessions table (nothing writes kind: "ranger" to agent_sessions any more). Same situation as Explorer, which the file comment explicitly excludes. Ranger should get the same treatment (or a plugin read).

3. /agents gates a null-team request as free; /qa-agent and the sidebar lock skip the gate when team is null — the three surfaces the e2e test expects to agree now diverge (agents/page.tsx:57 vs qa-agent/page.tsx:44). Deliberate per the comment, but mirror it in /qa-agent or expect confusion.

4. agents-console-client.tsx is "use client" with no hooks/handlers; timeAgo() at render time can hydration-mismatch on a minute boundary. Dropping the directive makes it a plain server component and removes the issue.
activeStep in core also matches failed steps; the QA header's copy it claims to mirror doesn't — the two "must produce the same sentence" already disagree. Promoting the three helpers to a shared lib would fix it for good.

5. Dead surface: fleetLabel is unused, Escalation.kind is never read, and the needs_input re-filter in escalationsFrom duplicates what getQaConsoleQueueRows already does.

ewyct:
Thanks for the thorough pass. Addressed in las-team/lastest@6913c8e7 on `feat/healer-agent`, the new top of the stack (above #126). Item by item:

**Review**

1. **`blocked` unreachable.** `deriveState` now checks for a `waiting_user` step first, on `active` or `paused`, and only then falls through to `paused` / `working`. The unit fixtures were changed to the combination the real writers produce (`status: "paused"` + `waiting_user`), plus a test pinning that both `active`+gate and `paused`+gate read as blocked.
2. **e2e selector.** Already fixed further up the stack (#123): both gating tests locate `nav a[href='/agents']` and assert the lock badge on the Agents item.
3. **Escalation "waiting since".** `FleetRow` gained `blockedSince`, sourced from the waiting step's `AgentStepState.startedAt` (stamped by `setStepActive` right before the gate), and `escalationsFrom` sorts/renders on `blockedSince ?? startedAt`. Test added for the 09:00-start / 12:00-gate vs 10:30-task ordering you described.

**Follow-up**

1. **Raw float run-minutes.** Console now uses `computeRunUsageProjection().usedPct` for the bar and new shared `formatRunMinutes` / `formatRunCount` helpers in `src/lib/billing/run-usage.ts`; the settings card uses the same helpers instead of its local `formatNumber`, so the two render identically.
2. **Idle Ranger row.** Ranger, Play and QuickStart are being dropped from the roster entirely in a separate change (they are onboarding / one-shot flows, not agents that work a repo), so not patched here.
3. **Null-team gate divergence.** `/qa-agent`, both `/triage-agent` pages and the sidebar lock now all evaluate `hasQaAgentAccess(team?.plan ?? "free", …)`, the same verdict as `/agents`.
4. **`"use client"` + drifted helpers.** The console is now a plain server component (`agents-console.tsx`, renamed since it is no longer a `*-client`), so `timeAgo()` runs once and cannot mismatch. `activeStep` / `sessionNarration` / `sessionProgress` moved to a new pure lib `@lastest/agent-steps` (`libs/`, third tier per core-scope §3); `fleet.ts`, the QA header and the QA phase timeline all import from it, so the "same sentence" claim is now enforced rather than asserted. Failed steps count as current everywhere, matching what the timelines highlight.
5. **Dead surface.** Removed `fleetLabel`, `Escalation.kind`, `EscalatableTaskLike.status` and the `needs_input` re-filter; the parameter is now `needsInput` and documented as pre-filtered by `getQaConsoleQueue`.

mtolmacs (approved):

PR #114, opened 2026-08-27
feat(agents): fold the Explorer into the console roster

ewyct:
**Stack: 2 of 2.** Base is `feat/agents-console` (#113) — review that one first. Split out because it touches the explorer plugin as well as core, and those ship as separate PRs per `core-scope.md` / RFC §7.2.

## What this adds

Explorer runs appear on the `/agents` roster beside core's agents, and Explorer loses its own sidebar entry — `/agents` becomes the single door to agent work. The Explorer page keeps its route and gains the same breadcrumb the QA agent page has.

## The boundary, and why it is shaped this way

`explorer_sessions` is the plugin's table; core cannot select from it. So the plugin **answers the question instead of exposing the table**:

- `readLiveExplorerSession()` (`plugins/explorer/src/fleet.ts`) returns a **projection**, not `ExplorerSession`
- core reaches it through `src/lib/core/explorer-reads.ts`, the mirror of the existing `share-reads.ts`

The projection is the part worth a second look. The session row carries an **encrypted login password** in `metadata`; the console has no business holding one. A narrow return type makes that structural rather than a rule someone has to remember, and it keeps core free of explorer's step/metadata types so the plugin can change them without a core PR.

Authorization is `contextFor()`'s, exactly as for every action in that package — there is no unscoped handle to reach for, so this cannot read a repo the caller cannot see.

## Notes for the reviewer

- **`rowFromExplorer` lands the projection on the same `FleetRow`** the core sessions produce, so the roster's rules (blocked before working; a blocked agent still holds its browser) apply without a special case.
- **`awaitingUser` is always `false` today** — explorer's pipeline has no human gate. It is in the projection so that adding one later needs no core change, and there is a test pinning that wiring so the field cannot rot into a lie.
- **The breadcrumb crosses as a `ReactNode` prop**, not an import — a plugin cannot import a core component, so it uses the seam `upgradeGate` and `noRepository` already use.
- **The read fails soft to `null`.** A roster missing one row beats a 500 on the console, so the failure is swallowed once in the read port rather than at each call site. Say the word if you would rather it surface.

## Verification

- `pnpm exec vitest run` — 128 files, 1921 passing (5 new tests covering the explorer row)
- `pnpm exec tsc --noEmit` — clean
- `pnpm build` — compiled successfully

### DIFF
--- a/plugins/explorer/src/data/queries.ts
+++ b/plugins/explorer/src/data/queries.ts
@@ -188,6 +188,33 @@ export async function getActiveSession(
   return row ? decryptSession(ctx, row) : undefined;
 }
 
+/**
+ * The repo's live run for the Agents console — `active` **or** `paused`.
+ *
+ * Distinct from `getActiveSession`, which is what the Explorer page drives and
+ * must stay `active`-only: resuming, stepping and cancelling all act on a run
+ * that is actually running. The roster has the opposite need — a paused run is
+ * still a roster row, and showing it as "idle" is exactly the under-report the
+ * console exists to prevent.
+ */
+export async function getLiveSession(
+  ctx: Ctx,
+  repositoryId: string,
+): Promise<ExplorerSession | undefined> {
+  const [row] = await ctx.db
+    .select()
+    .from(explorerSessions)
+    .where(
+      and(
+        eq(explorerSessions.repositoryId, repositoryId),
+        inArray(explorerSessions.status, ["active", "paused"]),
+      ),
+    )
+    .orderBy(desc(explorerSessions.createdAt))
+    .limit(1);
+  return row ? decryptSession(ctx, row) : undefined;
+}
+
 export async function getLatestSession(
   ctx: Ctx,
   repositoryId: string,
--- a/plugins/explorer/src/fleet.test.ts
+++ b/plugins/explorer/src/fleet.test.ts
@@ -1,0 +1,85 @@
+import { describe, expect, it } from "vitest";
+
+import { projectFleetSession } from "./fleet";
+import type { ExplorerStepState } from "./types";
+
+/**
+ * The plugin half of the Agents-console projection.
+ *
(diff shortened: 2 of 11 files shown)


ewyct:

| -1 | 0 | 30 | 0 | 10 |

[View Build](http://localhost:3000/builds/6448659e-704a-43b7-8661-9ff57f0e4b72)

---

ewyct:
Checked against the #97 pattern list — the "computed/declared but never wired" one lands here.

* **The `paused` half of the projection is unreachable** [`plugins/explorer/src/fleet.ts`]: `readLiveExplorerSession` gets its row from `q.getActiveSession`, which filters to `status = 'active'`. So the `status !== "active" && status !== "paused"` narrowing can only ever pass `active`, and everything downstream that handles `paused` is dead: the `"active" | "paused"` union on `ExplorerFleetSession`, the `paused` branch in `rowFromExplorer`, and the `frees the browser when the run is paused` test — which passes because it calls `rowFromExplorer` directly and never goes through the read. A paused Explorer shows on the roster as **idle**, so the console under-reports held browsers exactly where it matters. Either widen the query or drop `paused` from the type.
* **`awaitingUser` is hardcoded `false`** and documented as a forward seam. Fine — but the test that "pins the wiring" also tests `rowFromExplorer` in isolation, so nothing pins the plugin side. The day Explorer grows a gate, this field can still ship `false` with every test green.
* **`getLiveExplorerSession` swallows every error to `null`** [`src/lib/core/explorer-reads.ts`]. An authorization failure, a DB outage and "no session running" are then indistinguishable, and the roster confidently shows an idle Explorer while one is running. The fallback is right; log the error before returning null.
* `rowFromExplorer` re-declares the projection as an inline structural type instead of importing `ExplorerFleetSession` from `@/lib/core/explorer-reads`. Two shapes that must stay byte-identical with nothing enforcing it — add a field to the projection and this silently drops it.

mtolmacs (approved):

PR #116, opened 2026-08-27
feat(pharma): segment fork in onboarding + the regulated profile

ewyct:
Last PR in the pharma stack (#97 → #113 → #114 → this).

Onboarding now opens on a segment question — **Pharma & life sciences** or
**Custom app testing** — instead of jumping straight to "how do you want to build
tests?". The custom path is unchanged. The pharma path creates a project with the
Veeva Vault and Salesforce release-regression suites already in it, turns on the
regulated profile for the team, and lands on `/tests`.

Companion to `docs/pharma-restricted-scope.md`, which this implements §3 of.

## The profile

`src/lib/segment/regulated.ts` is the single place every restriction reads from,
rather than a `team.regulatedMode &&` scattered across call sites. It encodes
three rules: nothing probabilistic produces a verdict, nothing leaves the tenant
without an authenticated identity attached, and nothing on screen suggests the
tool is a game.

**Eval defaults for a validated system.** `text` and `dom` are raised to
`enforce` — the load-bearing change, because a template change that drops
"Meaning" from a 21 CFR Part 11 §11.50 signature manifestation is a compliance
finding, and the product default (`log`) lets it pass green. `network` and
`console` drop to `log`, since Vault and Lightning poll noisily enough to bury
the real findings; ratchet after one clean cycle. `perf`, `storage`, `a11y` and
`design` go off.

**UI.** Leaderboard, Agents, gamification, early-adopter and repo/CI integration
cards are hidden. Public share links are refused server-side in
`publishBuildShare` — not merely hidden. An anonymous `/r/<slug>` URL serving
screenshots of a validated system is a data-protection problem, and hiding the
button that mints one is not a control. Everything else here is merchandising.

## Step numbering

Steps 0 (fork) and 6 (pharma setup) sit *outside* the original 1–5 numbering
rather than renumbering it, so `?step=` deep links
(`connectGithub("/onboarding?step=2")`), the Esc-to-skip range and every existing
step body keep meaning exactly what they did. A deep link into 1–5 implies the
custom segment rather than re-asking the fork and dropping the user's OAuth
round-trip.

## Verification

Against a running app with a real signed-up user, not only unit tests:

| | regulated OFF | ON |
|---|---|---|
| Leaderboard / Arcade / Agents | visible | hidden |
| Coverage / Verify / Tests / Runs / Setup | visible | visible |
| GitHub / GitLab / Gamification / Early-adopter cards | visible | hidden |
| Regulated-mode toggle | visible | visible |

Gamification was explicitly **on** at team level in that run, so it exercises the
override rather than a default. The seed's insert path is covered against a real
database (quarantine flags, version rows, check modes, idempotence). Full unit
suite 1942 passed, typecheck clean, `pnpm arch` 0 violations.

## Two things to look at

**Both seeded tests arrive quarantined, and the setup screen says so.** They
cannot run: there is no per-repo secret store, so `process.env.VAULT_USER`
resolves against the embedded browser's own environment
(`docs/pharma-restricted-scope.md` §2.1). Seeding them green-able would be a lie.

**`toggleRegulatedMode` does not re-apply check modes to existing projects.**
Those are per-repo defaults a validation lead may have tuned, and rewriting them
from a settings switch would overwrite a deliberate choice. The toggle's help
text says so. Turning it on for an existing team changes UI and sharing, not eval
defaults.

### DIFF
--- a/packages/db/src/schema/identity.ts
+++ b/packages/db/src/schema/identity.ts
@@ -86,6 +86,21 @@ export const teams = pgTable("teams", {
   /** Verify phase (v1.14+) — when true, /verify is the primary surface and
    *  appears as the first sidebar entry. /run and /review are demoted. */
   verifyPhaseEnabled: boolean("verify_phase_enabled").default(true),
+  /**
+   * Regulated (pharma / life-sciences) segment profile — set by the Pharma
+   * onboarding fork, and the single flag every restriction in
+   * `src/lib/segment/regulated.ts` reads.
+   *
+   * It encodes three rules: nothing probabilistic produces a verdict, nothing
+   * leaves the tenant without an authenticated identity attached, and nothing
+   * on screen suggests the tool is a game. See
+   * `docs/pharma-restricted-scope.md` §3.
+   *
+   * Deliberately a *team* flag rather than a per-repo one: it governs
+   * tenant-wide surfaces (public share links, the leaderboard, billing) that
+   * have no repository in scope to key off.
+   */
+  regulatedMode: boolean("regulated_mode").default(false),
   storageQuotaBytes: bigint("storage_quota_bytes", { mode: "number" }).default(
     10737418240,
   ), // 10 GB
--- a/plugins/share/src/actions.ts
+++ b/plugins/share/src/actions.ts
@@ -58,7 +58,7 @@ export async function publishBuildShare(
   // validated GxP system, so hiding the button that mints one is not a
   // control — this is. See `docs/pharma-restricted-scope.md` §3.3.
   const ownerFlags = await host.getOwnerTeamFlags(info.repositoryId);
-  if (ownerFlags?.regulatedMode) {
+  if (ownerFlags && !ownerFlags.sharingPermitted) {
     throw new Error(
       "Public share links are disabled for this team. Regulated mode is on, and a public link would serve run screenshots to anyone holding the URL.",
     );
--- a/plugins/share/src/actions.ts
+++ b/plugins/share/src/actions.ts
@@ -53,6 +53,17 @@ export async function publishBuildShare(
   });
   const actor = requireActor(ctx);
 
+  // Regulated (pharma) profile: refuse outright. A `/r/<slug>` link is an
+  // anonymous, unauthenticated URL serving screenshots of what may be a
+  // validated GxP system, so hiding the button that mints one is not a
+  // control — this is. See `docs/pharma-restricted-scope.md` §3.3.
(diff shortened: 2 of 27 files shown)


ewyct:

| 0 | 0 | 0 | 0 | 39 |

[View Build](http://localhost:3000/builds/ad6e5784-a7a7-4960-b266-482586e44c74)

---

ewyct:
Reviewed against the #97 patterns. The dominant one here is **declared-but-never-wired**, and one of the unwired things is described in the code as the profile's only hard control.

* **Turning regulated mode on does not revoke shares that already exist.** `publishBuildShare` refuses to mint new ones, but `/r/<slug>` never consults `regulatedMode` — every previously minted link keeps serving run screenshots to anyone holding the URL, which is the exact scenario `regulated.ts` and the `ShareHost` docstring say is unacceptable. The toast ("Regulated mode on — public share links are now refused") reads as "the links are gone". Either revoke live shares inside `toggleRegulatedMode`, or make the public page 404 when the owning team is regulated.
* **`REGULATED_LOCKED_POLICY` has no callers.** `autoApprove: false`, `confirmOnGreen: false`, `aiDiffing: false`, `builtInAi: false` are documented as "forced, not defaulted" and "rendered as visible-but-disabled controls". `git grep` finds the symbol only in its own file. Nothing is forced and nothing is disabled — a regulated tenant can still turn auto-approval and AI diffing back on.
* **`REGULATED_HIDDEN_SETTINGS` has no callers either.** `settings/page.tsx` hardcodes `!regulated &&` around each card instead of consulting the set, so the set is documentation that looks like a mechanism: adding an id to it does nothing.
* **`isSharingPermitted()` has no production caller, but has tests.** `plugins/share/src/actions.ts` re-implements it as `ownerFlags?.regulatedMode`. The green test on the intended control point is the misleading part — it asserts a function nothing calls.
* **`REGULATED_CHECK_MODES` only reaches the DB through `seedPharmaSuite`.** Its docstring says "seeded for a repo created under the regulated profile", and `regulated-mode.ts` says "the profile seeds them at project creation instead" — but a repo created any other way (`createLocalRepo` from `/tests`) while the team is regulated gets product defaults. Given `text: "enforce"` is called out as the single most important change in that table, this gap is load-bearing, not cosmetic.
* **`startPharmaOnboarding` has no compensation path.** If `createLocalRepo` or `seedPharmaSuite` throws, the team is left `regulatedMode: true` with `onboardingPath` already rewritten to `"manual"` and no repo — outside the fork, with no route back into it. The ordering comment explains why the flag goes first, which is fine; the missing piece is the rollback of `onboardingPath`.

mtolmacs (approved):

PR #117, opened 2026-08-27
feat(credentials): per-repo secret store with run-time injection

ewyct:
Implements `docs/credentials-plan.md`. Sits at the end of the pharma stack, on top of #116.

Closes the blocker named in `docs/pharma-restricted-scope.md` §2.1: there was no per-repo secret store, so `process.env.VAULT_USER` inside a test resolved against the embedded browser's own process environment and the seeded Vault/Salesforce tests could never run. Not a pharma-only feature — every customer with a login wall behind their app needs it.

## The constraint that shapes everything

**A credential must not travel the variable-substitution path.** Every other variable is textually substituted into the test source before dispatch, and three things then happen to that string, all disqualifying for a password:

1. it is hashed into `codeHash` — a rotated password would change the hash, and a rotation is not a code change;
2. every resolved assign-mode value is persisted to `test_results.assignedVariables` as plaintext jsonb, once per run, forever;
3. the Variables tab renders those values back as its "Last run" column.

So credentials get their own channel: a `credentials` parameter injected at the four `AsyncFunction` sites the EB runs bodies at. No token grammar to learn, autocompletes in the editor, nothing to substitute.

```js
await page.getByLabel(/username/i).fill(credentials.vaultAdmin.username);
await page.getByLabel(/password/i).fill(credentials.vaultAdmin.password);
```

## What's here

- **`repo_credentials`** — one row per credential *set*, not per field: a username and a password belong together, and a flat bag gives the UI no way to say which password goes with which username. Secret values encrypted with the existing `crypto-fields` primitive (the arrangement `setup_configs.authConfig` already runs), non-secret fields in the clear so the list renders without a decrypt per row. No `environmentId` yet — that's gap-analysis B2.
- **Write-only values.** No read path returns a secret's plaintext; the query layer splits `listCredentials()` (masked, what the UI gets) from `getCredentialsForRun()` (plaintext, called at dispatch). Recommending against a reveal control for v1 as the plan argues: there's no audit log to record a reveal against yet (P1).
- **Decrypt-at-dispatch** in the executor, immediately before `createMessage(...)`, never cached on a request-scoped object.
- **Redaction.** The EB is the only process holding plaintext, so it scrubs its own output: log lines, soft errors, assertion messages, console errors/entries, thrown error messages and stacks, DOM snapshots, and captured page text.
- **Setup → Credentials** tab: list, editor dialog, copy-the-reference on every field. **Variables tab** gains a read-only reflection — one table, two readers, no mirroring.
- **Pharma seed** re-pointed from `process.env.VAULT_USER` to `credentials.vault.username`.

## Two deviations from the plan, both flagged

- **The seeded tests stay quarantined** (plan item 9 says un-quarantine). The *platform* block is gone, but the target URL is still a placeholder — pointing a seeded test at a real Vault would be pointing it at somebody else's tenant — and the credentials don't exist until the consultant adds them. Un-quarantining would guarantee a red first build, which the seed's own header calls out as the thing to avoid. One toggle for the customer once both are filled in.
- **§6.3 needed no code.** All five reference shapes already survive `redactCodeSecrets` untouched (every pass needs a quoted value; a reference is an expression). Added the documented invariant and a regression test rather than restoration logic that would do nothing.

`executeSetupViaRunner`'s two AI-authored-script callers (`play-agent`, `quickstart`) are deliberately left unwired, with a note at each: both already carry the one login the user handed the agent, and passing the repo's whole credential set into AI-generated code is the exact risk plan §9 raises.

## Verification

- `pnpm test` — 1973 passed, 132 files
- `pnpm test:integration src/lib/db/queries/credentials.integration.test.ts` — 9 passed against the real dev Postgres, including a raw-column read proving ciphertext is what's stored
- `pnpm lint` — 0 errors; app + EB + protocol + db typecheck clean
- `pnpm db:push` applied

New tests: crypto round-trip and non-double-encryption; the scrubber (longest-first masking, regex metacharacters, secret-key classification, error/DOM/text scrubbing, freeze semantics); the separation invariant (a `credentials.*` reference survives all three substitution passes, produces no `assignedVariables` entry, and leaves dispatched source byte-identical across a rotation); and the seed's shift from `process.env` to the store.

## Open question carried forward

Plan §9: repo-wide `credentials` means any test — including one an AI agent authored — can read every credential the repo holds. Shipped repo-wide per §3; the smallest fix if that lands wrong is `usableByTests: 'all' | testId[]` on the credential rather than a new table.


ewyct:

| -18 | 0 | 28 | 0 | 29 |

> **vs Main drift:** 25 tests have drifted from the main baseline

[View Build](http://localhost:3000/builds/28830107-dc13-48e2-b4fd-fed8c95fa55f)

---

ewyct:
Reviewed against the #97 patterns. Worth saying first: **`guardCredential()` is the right shape** — it loads the row, re-derives `repositoryId` from it, and only then calls `requireRepoCapability`. That is precisely the check the coverage cell/dimension actions were missing on #97, so this module is not exposed to that IDOR.

The redaction layer is where the gaps are.

* **Network capture is not scrubbed at all.** `EmbeddedNetworkRequest` carries `postData`, `responseBody`, `requestHeaders` and `responseHeaders`; none of them pass through `credScrub`. A login form submit puts the password in `postData` verbatim, an API login puts it in the request body, and a bearer token comes back in `responseHeaders` — all of it persisted for the `network` check layer. `credential-redaction.ts` opens by enumerating "three consumers" (log lines, error messages, DOM snapshots) and the single most certain carrier of a plaintext credential is not one of them. Everything else in this PR is careful; this one omission undoes a good part of it.
* **The scrubber ignores the field's own `secret` flag and re-guesses from the key name.** `CredentialField.secret` is authoritative — it decides encryption at rest, and `mergeSecretFields` trusts it — but `RunCredentials` flattens to `Record<string, Record<string, string>>` on the wire, so the EB falls back to `defaultIsSecretKey` keyword matching. A field the user explicitly marked secret and named `passphrase`, `vaultCode` or `clientAssertion` matches none of `SECRET_KEY_HINTS`: encrypted at rest, printed in the clear. Carry the secret key list on the payload instead of re-deriving it; the declared flag is already there and is being discarded on the wire (the #97 "declared but never wired" shape).
* `MIN_SCRUBBABLE_LENGTH = 4` silently excludes short secrets while `pin` and `otp` are in the hint list. A 3-character value is never masked. Worth stating that rule where a user sets a PIN rather than leaving it implicit in the EB.
* **`resolveRunCredentials` returns `undefined` on any failure**, so a bad `ENCRYPTION_KEY` or one corrupt row produces a run that fails on `credentials.x is undefined` with no indication of why. The `log.warn` is server-side only — nothing surfaces on the run. Failing the run with "credentials could not be decrypted" would be a better trade than a confusing green-path failure.
* `lastStampedAt` is never evicted — one entry per repo, retained for the process lifetime. Bounded by repo count, but it is a permanent map in a long-lived process.
* Noting without objection: every test receives **every** credential the repo holds. It is argued in `docs/credentials-plan.md` §3, so this is just flagging the consequence — one hostile or third-party test body reads the entire store, and per-test scoping is the mitigation if that lands badly.

mtolmacs (dismissed):

mtolmacs (approved):

PR #119, opened 2026-08-27
feat(webmcp): expose Lastest to browser AI agents via document.modelContext

ewyct:
Stacked on #117 (`feat/credentials`).

WebMCP lets a page hand typed tools to whatever AI agent the user is running. Chrome shipped the origin trial in 149; on 2026-08-25 OpenAI shipped it in the **ChatGPT desktop browser, ChatGPT Work and Codex** as "site tools". This registers Lastest's tools there, so an agent can review a build or approve diffs on the page the user is already looking at, using their signed-in session.

Research and the remaining slices: `docs/design/webmcp.md`.

## A bridge, not a third tool surface

We already define the tools once (`packages/mcp-server`, 29 `server.tool()` calls) and serve them twice (stdio CLI, `/api/mcp` over Streamable HTTP). This adds a third consumer without a third definition:

```
page → document.modelContext.registerTool(...)
         execute() → POST /api/mcp/session {op:"call"}
                       → createServer() over InMemoryTransport
                         → /api/v1/* → existing guards
```

- `/api/mcp/session` is the cookie-authed sibling of `/api/mcp`, exposing `tools/list` / `tools/call` as plain JSON — **no MCP SDK in the client bundle**.
- `LastestClient` gained an optional `apiKey` plus `extraHeaders`, so the route forwards the caller's session cookie to `/api/v1/*` (whose `getCurrentSession()` is cookie-first) rather than minting an API key per page load. `requireTeamAccess` / `requireRepoAccess` / capabilities all apply unchanged: an agent reaches exactly what the signed-in user reaches.

## Safety

- **CSRF:** a cookie-authed JSON endpoint is a CSRF target, so the route demands both a same-origin `Origin`/`Sec-Fetch-Site` and a custom `x-lastest-webmcp` header (non-simple ⇒ always preflighted, and there is no CORS handler). Covered by tests.
- **Narrowing:** several MCP tools multiplex read and destructive actions behind one `action` enum (`lastest_test` covers list/get/update/delete). Registering them verbatim would hand an agent `action:"delete"` and wreck tool selection. Each of the 16 registry entries pins the action via `source.bind` and publishes a small hand-written schema; storage-state, setup-script, create-test, explorer/ranger/quickstart are never sourced at all.
- **Route ids come from the page**, never from the agent — it cannot re-target a tool at another build by passing an id.
- **Consent:** every mutation is `readOnlyHint: false` and goes through `requestUserInteraction()` (with an in-page confirm fallback for the polyfill path). ChatGPT runs its own safety review too; we do not depend on any client's policy.
- `registry.test.ts` lists the live MCP surface in-process and fails on drift or on any breach of the above.

## Scope shown to an agent

3 global (list projects, running jobs, job status) + up to 6 route-scoped: repo (builds, failing tests, QA summary, coverage, run tests) / build (review, diffs, change map, approve, reject, publish share) / test (get, heal). ~10 per page, matching the WebMCP guidance and OpenAI's own showcase site.

## Enabling

`WEBMCP_ENABLED=1` (server-side env, deliberately not `NEXT_PUBLIC_*` — those bake into the client bundle at build time). Off by default, and inert in any browser without `document.modelContext`.

## Tests

30 assertions across `src/lib/webmcp/{registry,arguments,model-context}.test.ts` and `src/app/api/mcp/session/route.test.ts`. Typecheck and eslint clean.


ewyct:

| 0 | 0 | 0 | 0 | 39 |

[View Build](http://localhost:3000/builds/3ccef572-9f44-484a-9169-465c63c3393e)

---

ewyct:
Reviewed against the #97 patterns. The CSRF work on `/api/mcp/session` is solid — custom header *and* same-origin, with the no-`Origin` case correctly falling through to `Sec-Fetch-Site`, and no CORS handler anywhere. The concern is the public route.

* **`/api/webmcp/share/[slug]` does the full build fetch on every request, unauthenticated and unrate-limited.** `getBuildRenderContext` (all diffs + all results + all step comparisons) runs before the `switch`, so `report_summary` — which needs about eight counters — pays for the entire build, and `force-dynamic` means nothing is cached. This is the same shape as the Coverage page fetch flagged on #97, except the caller does not need an account and can loop it. A share's contents do not change; this is cacheable on the slug, and the per-op work should be fetched per-op.
* **The route bypasses regulated mode (#116).** A team that flips `regulatedMode` can no longer mint shares, but their already-live ones just gained a structured, machine-readable extraction endpoint over the same data. Whatever revocation story #116 ends up with needs to cover this route too, not just `/r/<slug>`.
* No per-slug or per-IP throttle on a route that is deliberately anonymous and deliberately does real database work. Worth at least a cheap cap.
* `WEBMCP_SHARE_OPS` membership is checked before the share is loaded — good ordering, keeps an unknown tool name from costing a DB read.

mtolmacs (approved):

PR #120, opened 2026-08-27
feat(triage): Triage agent as the single classifier, with the Run Results screen

ewyct:
Stacked on #119 (`feat/webmcp`). Review that first.

Replaces the two independent AI passes that ran at build completion — the per-diff analyzer and the per-failed-test failure triage — with **one build-scoped Triage agent** that clusters failures by root cause rather than by functional area, writes a run-level narrative, and suggests a verdict per cluster.

Its UI is the **Run Results** screen, reached by drilling into the Triage row on `/agents` — the same pattern as the QA agent and the Explorer.

## Placement

Follows the coverage split, not the plugin pattern:

- `libs/triage-model/` — pure: clustering, risk, verdict suggestion, ranking. No DB, no clock, no AI client.
- `src/lib/triage/` — stateful: orchestration, persistence, session lifecycle, the LLM call.

Not a plugin: the agent's product is writing core-owned rows (`visual_diffs`, `test_results`, `step_comparisons`), which a plugin cannot import.

## Data

Four new tables — `triage_runs`, `triage_groups`, `triage_cases`, `triage_case_verdicts`.

Reviewer verdicts live in their **own** table keyed on `(buildId, testId, stepLabel)`, so re-triaging a build replaces the agent's groups and cases but never clobbers a human decision. `stepLabel` there is `NOT NULL DEFAULT ''` — Postgres treats NULLs as distinct in a unique index, which would otherwise let the same case be decided twice and break the upsert.

Also: `agent_sessions` gains kind `triage` + four steps; `aiSettings.triageAgentEnabled`.

**Needs `pnpm db:push` before this runs.**

## Replace, not break

The agent now writes the two classification columns the retired passes owned — `test_results.triage` and `visual_diffs.aiAnalysis` / `aiRecommendation` / `aiAnalysisStatus` — so the healer, RCA, the verify board, the MCP server and the build screen keep reading exactly what they read before. Every consumer was enumerated and checked.

A failed triage marks unclassified diffs `failed`, so the build screen's "AI analysis failed" affordance stays alive rather than going permanently dark.

**One behaviour delta worth a reviewer's eye:** `aiAnalysisStatus` is no longer ever `pending`/`running`, so the build screen's AI spinner never renders. Both states render as nothing, so this is invisible rather than broken — but it is shipped UI going quiet.

Deterministic clustering is the **floor**: with AI off or unavailable, the screen still populates from the pure pre-pass rather than going empty.

## Gating

Pro plan via `hasQaAgentAccess`, then the per-repo setting, then in-product AI. The setting **cannot be enabled** by a team that fails the plan gate. Non-Pro builds behave exactly as they do today. Self-hosted lifts the plan check, as `hasQaAgentAccess` already does.

## Scope held deliberately

- No global "resolve all" — per-cluster bulk only.
- **No assignment feature.** Ownership is a filed or linked GitHub issue, now coverable at cluster level by one issue via the existing verify issue system.
- The failure reel is a **queue** of the individual recordings, not a concatenated video. No clip trimming.

## Verification

`tsc --noEmit` clean · `pnpm lint` 0 errors · `pnpm format:check` clean · `pnpm arch` 0 violations · 2045 tests passing across 139 files (34 new in the model package).

Not verified in a browser — no DB migration has been applied yet.


ewyct:
Ez implementálja a TestReportInterfaceProto tervedet.

ewyct:

| 0 | 0 | 0 | 0 | 39 |

[View Build](http://localhost:3000/builds/0db043b4-de6f-4b7b-b059-1dc262f6baa6)

---

ewyct:
Reviewed against the #97 patterns. Two things this PR gets right and should stay that way: `recordTriageVerdict` explicitly nulls a case whose `buildId` does not match the authorized build, and `retryFlakyCases` filters the incoming `testIds` down to ones actually owned by the build's repo. Those are the checks the coverage mutations on #97 were missing.

* **`runTriageForBuild` forces a re-run by default**: `force: opts?.force ?? true`. The no-argument call — which is what the UI does — always re-runs the whole classifier. For an AI-backed pass that is real spend on every click, and the option cannot express its own default (passing `force: false` works, omitting it does not). If forcing is genuinely the intent for the UI path, pass it explicitly at the call site and default the option to `false`.
* **`{ ok: true, error: … }` is a contradictory result shape.** `recordTriageVerdict` and `recordTriageGroupVerdict` return `ok: true` with `error` set when `applyVerdictSideEffects` fails — so a GitHub issue that never got filed, or a baseline that never got approved, reads as success to any caller doing `if (!res.ok)`. The verdict *is* persisted, so `ok: true` is defensible, but it needs a distinct field (`sideEffectError`) rather than reusing `error`.
* `recordTriageVerdict` never checks that `input.testId` belongs to the authorized build's repo before writing a verdict row keyed by it. The build is authorized so the blast radius stays inside the tenant, but it lets rows accumulate against arbitrary ids that no case will ever match.
* `applyVerdictSideEffects` is documented as best-effort and swallows into a returned string — consistent with the above, and fine once the result shape is fixed.
* Confirmed the `onConflictDoUpdate` target on `triage_case_verdicts` is backed by `uniqueIndex("idx_triage_verdicts_case")` — no `42P10` risk there.

mtolmacs (approved):

PR #121, opened 2026-08-27
feat(qa-agent): rebuild the QA agent console on the new bento direction

ewyct:
Stacked on #120 (`feat/triage-agent`) — review that one first.

## What

The QA agent screen (`/qa-agent`, reached from Agents) was a vertical stack of cards. This reshapes it into the balanced-quad bento from the Claude Design direction (`QA Agent - Ongoing Mode.dc.html`):

- **status strip** — identity, state, narration, pipeline %
- **pipeline strip** — Preflight → … → Summary, full width
- **coverage matrix** beside four live quadrants: Doing now · Coverage · Up next · Watching
- **direction queue** beside **live activity**

Setup form, plan review, coverage detail, triggers and run history keep their place below the bento.

The design system colours were already in `globals.css`, so this is a layout/composition change, not a token change.

## Real data, no mock tiles

- `QaPipelineStrip` reads `session.steps` — live run, else newest settled run, else the pipeline shape with all phases pending. Per-phase detail lines are read from the payload each step wrote in `setStepCompleted`; a phase that reported no count renders none.
- Phase definitions moved to `src/lib/qa-agent/phases.ts` so the server action and the screen share one list.
- **Up next** counts down to `qa_agent_triggers.nextRunAt`, falling back to queue depth. **Watching** renders the configured triggers (this replaces the header's `triggerSummary` line).
- **Live activity** is seeded server-side — `getRecentActivityEvents` gained a `repositoryId` filter, since the SSE feed only carries events from connect time and the tile would otherwise always open empty.

## Refactors rather than duplicates

- `QaCoverageMatrix` extracted from `QaSummaryPanel`; the panel takes `showOverview={false}` and keeps the branch/group/journey detail.
- `PhaseTimeline` gains `detailOnly` so the active-step detail renders under the strip without repeating the phases (run history still uses the full form).

## Verification

`tsc --noEmit`, `eslint`, `pnpm build`, and `vitest run src/lib/qa-agent src/server/actions/qa-agent.integration.test.ts` (78 passing) all clean.

### DIFF
--- a/plugins/qa-agent/src/actions.ts
+++ b/plugins/qa-agent/src/actions.ts
@@ -97,6 +97,7 @@ import {
 } from "./domain/task-triage";
 import type { QaActivityEvent, QaAgentHost, QaExistingAuthSetup } from "./host";
 import { qaAgentPlugin } from "./index";
+import { QA_PHASES } from "./phases";
 import type { QaAgentTask } from "./schema";
 import type {
   QaAgentRole,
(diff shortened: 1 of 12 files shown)


ewyct:

| 0 | 0 | 0 | 0 | 39 |

[View Build](http://localhost:3000/builds/91d9aba4-6299-4932-a5eb-4ece3a5effbe)

---

ewyct:
Reviewed against the #97 patterns — this one is largely a UI rebuild and comes out clean on the cross-tenant and unbounded-work fronts. `getRecentActivityEvents(teamId, { repositoryId, limit: 20 })` filters in SQL rather than in JS, which is the right call.

Smaller things:

* **`QA_PHASES` is exported as a mutable array** and is now genuinely shared: the server action seeds `session.steps` from it and a client component renders the strip from it. A `.sort()` or `.reverse()` anywhere on the client silently reshapes what new sessions get seeded with. `as const` (or `readonly Array<…>`) makes the single-source-of-truth claim structural instead of conventional.
* The docstring on `phases.ts` points at `@/lib/qa-agent/phases`; the file is `plugins/qa-agent/src/phases.ts`. Same for the comment left behind in `actions.ts`.
* `QaFeedEvent.createdAt` widened to `string | Date | null` — worth confirming every consumer formats the null case rather than rendering `Invalid Date`. The type change is the easy half.

mtolmacs (approved):

PR #122, opened 2026-08-27
feat(mcp): OAuth 2.1 on the remote MCP endpoint, and an access-level tool policy

ewyct:
Stacked on `feat/qa-agent-bento`.

`/api/mcp` could only be reached with an API key, which means an agent platform can only use Lastest if a human first generates one and pastes it in. This makes the endpoint an OAuth-protected resource a client can connect to knowing nothing but its URL and — because that client is not the user — narrows what it can see.

Phases 1 and 2 of [`docs/architecture/agentexchange-listing-plan.md`](docs/architecture/agentexchange-listing-plan.md). Design notes: [`docs/architecture/remote-mcp-oauth.md`](docs/architecture/remote-mcp-oauth.md).

## OAuth 2.1

better-auth's `mcp` plugin supplies the authorization server (RFC 8414 metadata, RFC 7591 dynamic client registration, PKCE S256), backed by three new tables in `packages/db/src/schema/identity.ts`. A 401 from `/api/mcp` carries RFC 9728 `WWW-Authenticate`, and `/.well-known/oauth-{authorization-server,protected-resource}` serve discovery at the paths clients actually probe.

Two decisions worth reviewing:

- **Discovery uses the public origin.** `req.url` inside a route is the internal `127.0.0.1:3001` address `scripts/front-proxy.js` forwards to; advertising that as the issuer points every client at an unreachable host. The loopback `LastestClient` deliberately keeps `req.url`'s origin so tool calls don't round-trip back out through the proxy.
- **The advertised authorization endpoint is ours** (`/oauth/mcp/authorize`), which pins `prompt=consent` and forwards everything else. better-auth only shows consent when the client asks for it, and combined with anonymous client registration that would let an app registered seconds ago get a token silently.

## Access levels

An OAuth token is not an API key.

| Credential | Level | Tools |
| --- | --- | --- |
| API key | `full` | 29 |
| OAuth + `lastest:write` | `write` | 27 |
| OAuth, anything else | `read` | 16 |

The level is applied at tool **registration**, not inside handlers: a restricted caller's `lastest_test` advertises `action: "list" \| "get"` and nothing else. That matters more than a runtime check — an agent picks tools by reading schemas, and one that can see a forbidden action keeps trying it. Deletes, share revocation and publishing a public `/r/` link are unreachable over OAuth at any scope: not gated, absent. Table in `packages/mcp-server/src/policy.ts`, drift-guarded by `policy.test.ts` against the live surface.

**OAuth tokens are deliberately not accepted by `/api/v1`.** A `lastest:read` token that authenticates against REST directly would bypass the policy entirely, so `verifyBearerToken()` does not know about them. `/api/mcp` mints a 60-second HMAC loopback grant instead — server-created, server-spent, never in a response — same construction and fail-closed rule as the EB stream grant.

## Also in here

- **Bug fix:** `/api/mcp` was missing from `PUBLIC_PATHS`, so bearer-only clients were 307'd to `/login` and never saw the 401 that starts discovery.
- **Bug fix:** better-auth's `getMcpSession()` looks a token up without checking expiry — an expired token would work forever. `remote-auth.ts` checks it.
- `lastest-mcp --transport http` runs the same tool surface as a standalone server, for deployments that want the MCP endpoint on its own host/port.

## Verification

End to end against a local instance: discovery → dynamic client registration → login → forced consent → PKCE exchange (wrong verifier → 401) → `initialize` → scope-limited `tools/list` → `delete` rejected by the narrowed schema. Read/write/full observed as 16/27/29. Probe user and rows cleaned out of the dev DB.

`pnpm test` 144 files / 2075 tests pass, `pnpm lint` 0 errors, `pnpm build` clean with all new routes registered. `pnpm db:push` applied locally; the three tables are additive.

## Not done

Recorded in the plan doc rather than silently skipped:

- No per-connection rate limiting or audit logging (the route logs one debug line). Needed before an AgentExchange submission.
- Tool **responses** are not redacted — `redactSecrets()` only covers inbound params on their way to activity logs.
- No UI for reviewing or revoking OAuth connections. `getOAuthConnectionsForUser()` exists for it; Settings → Runners & API Access is the obvious home.
- Tool descriptions were not rewritten for AgentExchange's semantic search; that is listing-copy work and belongs with Phase 5.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -133,7 +133,7 @@ Visual regression testing platform: Next.js 16 App Router, PostgreSQL (Drizzle O
 - `packages/pool-service/` — EB pool service (separate process, `pnpm dev:pool`): k8s/process provisioning, pool caps, warm pool, EB reapers; app consumes `@lastest/pool-service/client`
 - `packages/eb-protocol/` — canonical wire protocol app ↔ runners (`@lastest/eb-protocol`): command/response messages, stream messages, persisted jsonb payload shapes (schema.ts re-exports these)
 - `packages/runner/` — remote runner CLI (npm package via tsup)
-- `packages/mcp-server/` — MCP server for AI agent integration (`@lastest/mcp-server`)
+- `packages/mcp-server/` — MCP server for AI agent integration (`@lastest/mcp-server`). Three transports off one tool surface: the stdio CLI, a standalone Streamable HTTP server (`--transport http`), and the app's own `/api/mcp` route. `src/policy.ts` holds the read/write/full access-level table applied at tool-registration time — see `docs/architecture/remote-mcp-oauth.md` for the OAuth 2.1 story (`/api/mcp` is an OAuth-protected resource; `/.well-known/oauth-*` + `/oauth/mcp/authorize` + `/oauth/consent` are the discovery/consent chain). Both `/api/mcp` and `/.well-known/` must stay in `PUBLIC_PATHS` in `src/proxy.ts`
 - `packages/embedded-browser/` — containerized browser with CDP live streaming
 - `packages/ocr-service/` — Tesseract OCR container, the ONLY OCR backend (no in-process Tesseract in the app); app-side facade in `src/lib/ocr/` requires `OCR_SERVICE_URL` — unset means OCR features are disabled. The service wakes on demand and auto-sleeps after idle
 - `packages/vscode-extension/` — VS Code extension (esbuild)
--- a/docs/architecture/agentexchange-listing-plan.md
+++ b/docs/architecture/agentexchange-listing-plan.md
@@ -1,0 +1,134 @@
+# Salesforce AgentExchange listing plan (Lastest)
+
+Researched 2026-08-27. Sources at bottom.
+
+## Can we register today?
+
+No — not without a human doing the legal/corporate steps, and not with the MCP server as it exists.
+
+- Registration is `partnersignup.salesforce.com` → **AgentExchange ISV Program Track**. Corporate application approval takes 1–3 weeks and requires a signed Partner Distribution Agreement (a legal signature, not something I can do).
+- A Partner Business Org is provisioned after approval (1–3 business days, free for year 1).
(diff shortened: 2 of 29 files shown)


ewyct:

| -3 | 0 | 28 | 0 | 14 |

> **vs Main drift:** 30 tests have drifted from the main baseline

[View Build](http://localhost:3000/builds/4ff2bfe2-7632-4b0b-83e0-d97fd613e48b)

---

ewyct:
Reviewed against the #97 patterns, plus the authorization surface this PR introduces. The loopback-grant design is good — HMAC over a payload, timing-safe compare *before* parsing, fail-closed with no `ENCRYPTION_KEY`, never returned to a client — and catching that better-auth's `getMcpSession` does not check `accessTokenExpiresAt` is a real save.

* **A token with no Lastest scope still gets the full read surface.** `accessLevelForScopes` returns `"read"` unless `lastest:write` is present, so a client that requested only `openid profile` — or a client whose scope string failed to parse, since `parseScopes` maps every falsy input to `[]` — reads every build, test, diff and coverage row the user can see. The module documents this as "OAuth, anything else → read", so it is deliberate, but it means the consent screen can grant data access for a scope the user reads as "sign in with Lastest". Refusing when neither `lastest:read` nor `lastest:write` is present would close it without changing the two-level model.
* **`/api/mcp` in `PUBLIC_PATHS` is a prefix match.** `src/proxy.ts` does `PUBLIC_PATHS.some((p) => pathname.startsWith(p))`, so this also exempts `/api/mcp/session` (added in #119) and every future `/api/mcp*` route. Nothing is open today — both routes authenticate themselves — but the default is now "public unless the route remembers", which is the wrong direction for this particular prefix. An exact-match entry, or a separate exact-match list, keeps the OAuth discovery flow working without the blanket.
* **The loopback grant confers `full` `/api/v1` power for 60s** and is bound to a user, not to the request or the tool being called. The scope restriction therefore lives entirely in the MCP tool layer, which the header states plainly — but the tests cover the *mapping* (`accessLevelForScopes` returns `"read"`). What is missing is a test asserting that a `read`-level caller actually cannot reach a write tool end-to-end. Given that property is the one the AgentExchange review cares about, it deserves the stronger test.
* `sessionId: \`mcp-oauth:${grant.c}\`` is a synthetic id that never corresponds to a `sessions` row. Worth checking nothing downstream treats `SessionData.sessionId` as a lookup key rather than a label.

mtolmacs (approved):

PR #123, opened 2026-08-27
feat(verify): retire /run and /builds into Verify, and unblock WebMCP

ewyct:
Stacked on #122.

`/verify` becomes the single execution surface. `/run` and `/builds/*` are redirect stubs, every client that backed them is deleted, and the `verifyPhaseEnabled` flag is gone from schema, UI and code. `docs/architecture/retire-run-build-pages.md` is updated to record what shipped against the plan.

## Carried into Verify

| From | Now |
|---|---|
| `/run` build-history list + graph | Lazy header drawer reusing `BuildGraphView` / `BuildSummaryCard` verbatim, one `getBuildHistory()` action (builds + branch heads). This was the piece with **no** Verify equivalent at all. |
| Run Tests / Smart Run / Comparison Run cards | Run split-button: smart (lazy `analyzeSmartRun` count + changed files), all, comparison vs base; compose chip; run-minute quota lock. The comparison *setting* moved to Settings → Repository — picking "Run comparison" is the decision the toggle used to persist. |
| Build page "Accept All Safe (AI)" | Unsorted-column action, identical write path to "Verify all" narrowed to AI-judged-safe cases (new `aiSafe` boolean on the slim diff). |
| Build page browser + RCA filters | `FilterPanel` rows (browsers limited to the engines the build actually ran). |
| Build page live browser view | Collapsible stream panel above the board while running. |
| Comparison-pair banner | Chip beside the branch picker. |
| Diff viewer step-label editor | Focus header. Renaming re-diffs against the matching baseline — a repair path, not decoration. |

Not carried over, deliberately: `/run`'s base-URL card, URL history, connection test and branch picker (all duplicates of `sidebar-quick-actions.tsx`).

## Redirects

`/run` → `/verify`; `/builds/<id>` → `/verify/<id>`; `/builds/<id>/diff/<diffId>` resolves `(testId, stepLabel)` → `step_comparisons.id` and lands on `?mode=focus&step=<id>`, falling back to the board. Minted URLs in GitHub issue bodies, Slack confirm-on-green and share records keep resolving. ~40 `revalidatePath` call sites and all in-app links repointed.

## WebMCP

On unless the deployment sets `WEBMCP_ENABLED=0`. The per-team toggle and `teams.web_mcp_enabled` are gone — they bought nothing over the three guards that actually hold (inert without `document.modelContext`, every mutation waits on our own consent dialog, tools act only with the signed-in user's permissions) while leaving the feature dark for every team by default.

Also fixes the `Tool already registered: lastest_list_projects` runtime overlay, which was two bugs:
1. The polyfill's `registerTool` is **async**, so duplicate-name and abort-on-dispose failures arrived as unhandled rejections the synchronous `try/catch` never saw.
2. The duplicate itself was StrictMode's double-effect — both passes awaited the polyfill install, and the stale pass's disposer ran a microtask *after* the fresh pass had claimed the same names. Cancellation is now an `AbortSignal` checked before registering.

Two regression tests added.

## Regulated profile

No longer hides Agents. Authoring tests is what that segment wants; what must never happen is an AI *verdict* in the execution record, and `REGULATED_LOCKED_POLICY` (`aiDiffing` / `builtInAi`) is what enforces that.

## Deploy note

**`pnpm db:push` required** — drops `teams.web_mcp_enabled` and `teams.verify_phase_enabled`.

## Checks

`tsc` clean · `eslint src e2e` 0 errors · `pnpm test` 144 files / 2077 passing · `next build` exit 0. Golden-path and webmcp e2e steps rewritten against the Verify UI (need a dev server to run).


ewyct:

| -5 | 0 | 31 | 0 | 13 |

[View Build](http://localhost:3000/builds/5cba9206-7f08-4f07-9f6b-9d14cdf7a173)

---

ewyct:
Reviewed against the #97 patterns. The redirect work is the right instinct — `/builds/<id>` and `/builds/<id>/diff/<diffId>` URLs are minted into GitHub issue bodies, Slack messages and share records, and keeping them resolving forever is correct. The concerns are about the migration, not the routing.

* **Destructive column drops, colliding with an add in the same stack.** This PR removes `teams.verify_phase_enabled` and `teams.web_mcp_enabled`, while #116 (lower in the stack) adds `teams.regulated_mode`. Deployed together that is an add + drop pair on one table, which is exactly the case `CLAUDE.md` warns about: `drizzle-kit push --force` hangs on the rename prompt, and a bare `db:push:raw` cannot distinguish a rename from a drop. Pre-apply the DROPs as explicit `ALTER TABLE`s, or land them in a separate step after the add.
* `web_mcp_enabled` was **added by #119 in this same stack** and is dropped here. If that is a deliberate flag graduation, the PR body should say so — from the diff alone an operator cannot tell a graduation from an accidental revert, and the column carries per-team opt-in state for a feature the docs describe as security-sensitive ("lets an agent the operator did not configure act with the signed-in user's permissions").
* Dropping `verify_phase_enabled` also removes the only way to put a team back on the old surface. That is presumably the point of retiring the pages, but it is a one-way door worth stating explicitly in the description.
* `plugins/rca/src/actions.ts` now calls `revalidatePath(\`/verify/${buildId}\`)` **twice in a row** — the rewritten `/builds/` line plus the pre-existing `/verify/` line. Harmless, but a leftover from the mechanical rewrite; worth a scan for others of the same shape.
* The `/builds/<id>/diff/<diffId>` redirect resolves through `(testId, stepLabel)` and falls back to the board when the lookup misses — good failure mode. Since the mapping is data-dependent, confirm this is a temporary redirect and not cached as permanent by any intermediary.
* `/api/webmcp/share/` added to `PUBLIC_PATHS` is consistent with #119's design; see my note there about that route's cost per anonymous call.

mtolmacs (approved):

PR #124, opened 2026-08-27
fix(ci): correct build progress math and unblock recorded tests

ewyct:
Stacked on #123.

The CI run on #121 printed `Progress: 66/39 tests complete` and finished 30 passed / 9 failed / 35 changes. Four separate defects, all reachable from that one build.

## Progress overshoot — `66/39`

`packages/runner` derived completion as `passed + failed + changes + flaky`. But `passedCount`/`failedCount` are counted **per test result** (`builds.ts:920`), while `changesDetected`/`flakyCount` are counted **per visual diff**, one per screenshot step (`builds.ts:962`). A passing test with three changed steps contributed four.

Test-level progress is `passed + failed`. The status API now serves that as `completedTests`; the runner prefers it and falls back to the sum for older servers. Every count adjustment (flaky retry, step-criteria override) moves a result *between* those two, so the invariant holds. The unit of each counter is now documented on `BuildSummary`.

## Double-counted setup failures

A `setup_failed` result was tallied into `failedCount` when recorded, then tallied **again** by the step-criteria override pass, whose guard only excluded `"failed"`.

## `urlMatch is not defined`

```
Assertion ecf346b637b0 failed: urlMatch is not defined (reached step 39/39)
```

The recorder emits `urlMatch(...)` for every click-triggered navigation — the prefix-regex that lets `/verify` tolerate redirecting to `/verify/<buildId>`. But `urlMatch` was only ever written into the wrapper preamble **as text**, while the executor hands every other helper (`expect`, `locateWithFallback`, `downloads`, …) in as a **function parameter**. A body arriving without that preamble threw at the first URL assertion.

The executor now supplies `buildUrl`/`urlMatch` when absent. Prepended, not injected as parameters: a body that *does* carry the preamble declares them as `const`, and a same-named parameter would make the whole compiled function a `SyntaxError`.

## Unbounded `networkidle` waits — the 240s timeouts

`networkidle` never fires against an app holding a long-lived connection open, and Lastest's own pages do exactly that via the `/api/runners/status` SSE stream (8s keepalive, `route.ts:113`). Each unbounded call burns a full navigation timeout; a recorded suite emits one per navigation, so an 11-navigation test blows past the executor's 240s cap and reports a timeout instead of the assertion that mattered.

A trailing `.catch(() => {})` does **not** make it safe — it swallows the rejection *after* waiting anyway, which is exactly what the failing tests do.

The generator no longer emits these, and the executor bounds the ones already stored in the DB (same mechanism as the existing `selectAll` rewrite). Explicit `{ timeout }` is left alone.

## Workflow headroom

Generated job timeout was `ceil(timeout/60000)` — identical to the runner's own poll timeout, so GitHub cancelled the job at the same instant the runner would have printed its diagnosis. Now +5 minutes.

## Deploy button

`plugins/ci` pinned `@lastest/runner@0.4.2`. Both generators and this repo's own workflow now pin `0.5.2`, so pressing **Deploy** on the GitHub Actions setup page writes a workflow with the fixed runner.

**`@lastest/runner@0.5.2` must be published to npm before the new YAML resolves.**

## Verification

- `pnpm test` — 145 files, 2083 passed
- `tsc --noEmit` clean across app, runner, embedded-browser
- `eslint` clean on all changed files
- New regression tests: `ensureUrlHelpers` (4 cases incl. the SyntaxError-avoidance path) and the bounded-networkidle codegen assertion

Not fixed here, flagged: `flakyCount` is written in **two units** — per-diff at `builds.ts:963` and per-test in the retry block at `builds.ts:1379`. Changing a stored column's meaning would reinterpret historical builds, so it needs its own decision.

### DIFF
--- a/.github/workflows/lastest.yml
+++ b/.github/workflows/lastest.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   visual-tests:
     runs-on: ubuntu-latest
-    timeout-minutes: 50
+    timeout-minutes: 55
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
@@ -19,7 +19,7 @@ jobs:
           LASTEST_TOKEN: ${{ secrets.LASTEST_TOKEN }}
           LASTEST_URL: ${{ secrets.LASTEST_URL }}
         run: |
-          npx @lastest/runner@0.4.2 trigger \
+          npx @lastest/runner@0.5.2 trigger \
             -t "$LASTEST_TOKEN" \
             -s "$LASTEST_URL" \
             --repo "las-team/lastest" \
--- a/libs/recording-codegen/src/event-to-code.test.ts
+++ b/libs/recording-codegen/src/event-to-code.test.ts
@@ -215,3 +215,38 @@ describe("eventsToCodeLines — cursor tracking + selectors", () => {
     expect(joined).toContain("modifiers: ['Shift']");
   });
 });
+
+describe("eventsToCodeLines — networkIdle assertion", () => {
+  it("bounds the wait and swallows its rejection", () => {
+    const out = lines([
+      {
+        type: "assertion",
+        timestamp: 1,
+        data: { assertionType: "networkIdle" },
+      },
+    ]);
+
+    // An unbounded waitForLoadState('networkidle') never resolves against an
+    // app holding a long-lived connection open (SSE / websocket / long-poll),
+    // so it burns the full navigation timeout on every call. A recorded suite
+    // emits one per navigation, which is enough to push a test past the
+    // executor's hard cap and report a timeout instead of a real result.
+    expect(out).toEqual([
+      "// Assertion: Verify no pending network requests",
+      "await page.waitForLoadState('networkidle', { timeout: 3000 }).catch(() => {});",
+    ]);
+  });
+
+  it("never emits an unbounded networkidle wait", () => {
+    const out = lines([
+      {
+        type: "assertion",
+        timestamp: 1,
+        data: { assertionType: "networkIdle" },
+      },
+    ]);
+    expect(out.join("\n")).not.toMatch(
+      /waitForLoadState\(\s*['"]networkidle['"]\s*\)/,
+    );
+  });
+});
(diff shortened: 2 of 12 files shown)


ewyct:

| 0 | 0 | 0 | 0 | 39 |

[View Build](http://localhost:3000/builds/9799581e-2dba-4021-bd68-6cf9eed590a7)

---

ewyct:

| 0 | 0 | 0 | 0 | 39 |

[View Build](http://localhost:3000/builds/7c12e1c8-7a0e-4443-8e65-c1fac68118d4)

---

ewyct:
Reviewed against the #97 patterns. The progress-math fix is right and the unit confusion is now documented where it belongs (on the `BuildSummary` fields themselves), which is the durable part. Two things about the `networkidle` change.

* **The rewrite turns a recorded assertion into one that can never fail.** The executor rewrites `page.waitForLoadState('networkidle')` to `…({ timeout: 3000 }).catch(() => {})`, and this happens *before* the assertion instrumentation that wraps `await page.waitForLoadState(...)` lines in `__assertion(id, …)`. Because the rejection is swallowed inside the wrapper, "Verify no pending network requests" now records `passed` unconditionally — on an app that never reaches idle, on a page that 500s, on anything. That is the #97 silently-discarded-work shape: the check still appears in the results, it just no longer checks. Bounding the wait is clearly correct; the question is whether the timeout should mark the assertion soft-failed rather than passed. At minimum it should not be reported as a pass.
* **Does `completedTests` ever reach `totalTests` when a result is neither passed nor failed?** `completedTests = passedCount + failedCount`, and the runner loops until it sees completion. If any terminal result status (skipped, cancelled) is tallied into neither counter, CI progress stalls one short and the poll runs to timeout instead of finishing. Worth an explicit assertion somewhere that every terminal status lands in exactly one of the two — the invariant is stated in the doc comment but nothing enforces it.
* `ensureUrlHelpers` prepends lines to the body before instrumentation, shifting every subsequent line number by one or two. Anything that maps a runtime stack frame back to the stored test source will now be off by that amount for these bodies.
* `declares()` matches on source text, so a body that merely mentions `const buildUrl` inside a string literal or comment suppresses the injection and reintroduces the exact `is not defined` failure this fixes. Narrow, but the failure mode is the one being fixed.
* `deploy.sh`: `npm view` returning non-zero for a network error is indistinguishable from "not published", so the republish guard fails open. And the `err` message interpolates `$(echo $pinned)` unquoted — if the two `RUNNER_VERSION` constants ever diverge, `sort -u` yields two lines and the message is mangled at exactly the moment it matters.

mtolmacs (approved):

PR #126, opened 2026-08-27
feat(environments): environments as a first-class object, with Vault and Salesforce connectors

ewyct:
Stacked on #124.

Answers "could we add a Salesforce and a Veeva Vault org connector under Integrations, multiple per type, populating the credential store" — yes, but connectors without environments only move the URL. So this lands gap-analysis **B2** as the spine and hangs the connectors off it.

## The problem

Three things already assumed Vault and Salesforce and could not reach each other:

- `src/lib/demo/pharma-seed.ts` seeds tests reading `credentials.vault.username`, then throws until a human hand-creates that exact handle with those exact field keys.
- `VaultProfiler` speaks the real Vault REST API, but `profileFromSutAction` took credentials **per call and never persisted them** — a comment there deferred storage to the environment model, and it had no UI caller at all.
- Baselines were keyed by **branch**. A Vault customer has PROD / UAT / prerelease 26R2, not branches, so two environments fought over one baseline row and every build flipped it back and forth.

## Design

```
environments            key, label, baseUrl, releaseLabel, refreshedAt
environment_variables   non-secret ids a sandbox refresh invalidates
sut_connectors          N per type per repo, bound to an environment,
                        pointing at a repo_credentials row
```

**One secret store.** Every secret a connector needs — Vault password, SF consumer secret, JWT private key — goes through `createCredential`. One encryption, masking, injection and EB-redaction path, all pre-existing.

**Test code never changes when the environment does.** Handles are unique per `(repo, environment)` and the resolver falls back to the repo-wide row, so `credentials.vault.password` is the UAT password on a UAT run and the PROD password on a PROD run. No substitution → no `codeHash` move → no invalidated baseline.

**Baselines stop fighting.** `environmentKey` on all eight baseline tables with the ordered NULL-fallback `baselines.dataCell` already proved: `(env, cell) → (env, ·) → (·, cell) → (·, ·)`. Environment is the **outer** loop deliberately — the environment decides what the page should look like; the cell only decides which data it showed.

## Adoption is additive

Every scoping column is nullable and every lookup falls back to NULL, so a repo that never opens the screen behaves exactly as before. `scripts/backfill-environments.ts` seeds one `prod` per repo — and deliberately does **not** claim existing baselines for it, since leaving them unscoped is what keeps them visible from every environment. Promote when you actually want them claimed.

## Also here

- **Promotion** (UAT sign-off → PROD) shares the PNG rather than copying it, and supersedes rather than deletes — a regretted promotion has to be reversible.
- **Refresh survival.** Nothing in a baseline's key changes across a sandbox refresh, so approvals carry over. What does change are record ids, which is why they live in `environment_variables` and are **injected** as `credentials.env.*` rather than substituted: a `{{env:docId}}` would land in `codeHash` and invalidate every baseline on the next refresh, destroying the property this exists to provide.
- **Two bugs found while wiring:** `deactivateBaselines` was environment-blind (approving in UAT would have dropped PROD's baseline), and `getBaselineByHash` needed the same scoping.
- **`scripts/migrate.js`** drops the old full `uq_repo_credentials_repo_name` before push. The index keeps its name and only gains a `WHERE` predicate, which drizzle-kit cannot detect — left alone it silently keeps enforcing repo-wide uniqueness and the second environment's `vault` login fails on insert. (Note `migrate.js` loads no `.env.local`, so its pre-push steps skip in local dev unless you export `DATABASE_URL`; Docker supplies it.)

## Connector research that shaped the auth methods

- **Salesforce disabled new Connected App creation in Spring '26**, and External Client Apps dropped the username-password grant entirely. API auth is therefore client-credentials or JWT-bearer only, and the form login survives purely as the browser path — modelled as `sf-ui-login` and marked *unverifiable* rather than given a fake green tick.
- Salesforce returns an `instance_url` that differs from the login URL; captured at verify time, never typed.
- Vault OAuth exchanges at Veeva's fixed `login.veevavault.com`, not the customer's Vault DNS.
- JWT `exp` must be within 5 minutes of Salesforce time, so `invalid_grant` is mapped to an explicit clock-skew message.

## Verification

- `pnpm test` — 148 files, **2111 passed**; `tsc` clean; lint 0 errors.
- Integration — **48 passed**, including 19 new cases pinning baseline precedence (with "a newer unscoped baseline must not win on timestamp") and per-environment credential resolution.
- e2e in a real browser — **8/8**: two environments, the same `vault` handle connected in both, ciphertext at rest, the secret absent from the DOM, blank-edit preservation, no orphan credential on disconnect.
- `scripts/backfill-environments.ts --dry-run` against the dev DB.

**Not done:** the connector clients are proven against stubbed fetches only — neither has been run against a real Veeva sandbox or Salesforce Developer org — and the backfill has not been run for real.


ewyct:

| -2 | 0 | 32 | 0 | 9 |

[View Build](http://localhost:3000/builds/c5d93a3c-6181-4441-b4cd-ce875f8d7113)

---

ewyct:
Reviewed against the #97 patterns. The SSRF story is handled properly — `connectorFetch` routes every outbound call through `safeOutboundFetch` with per-hop redirect revalidation, which is the right answer for a base URL the user types. But **the #97 IDOR pattern reappears here almost verbatim.**

* **`environmentId` is accepted from the client and never scoped to the repo.** `createConnector`, `updateConnector`, `createCredential` and `updateCredential` all authorize `repositoryId` (correctly) and then write `input.environmentId` straight through without ever checking the environment belongs to that repo. That is the same shape as the coverage cell/dimension mutations on #97: one thing authorized, a different raw id trusted. `environments.ts` in this very PR gets it right — `guardEnvironment(id)` re-derives `env.repositoryId` before the capability check, and `promoteBaselines` resolves keys via `getEnvironmentByKey(repositoryId, key)` — so the fix is one call in each of the four actions: load the environment and assert `env.repositoryId === repositoryId`.
  Concrete effects even without cross-tenant reads: a credential or connector can be bound to another tenant's environment id, and `connectorNameTaken(repoId, name, …, environmentId)` / `credentialNameTaken(…)` then probe uniqueness in a scope the caller does not own, which is an oracle for whether a given environment id exists.
* **`connectorFetch` silently drops a `Request`'s headers and body.** Its signature is `typeof fetch`, but when `input` is a `Request` it extracts `input.url` and forwards `init ?? {}` — which is `undefined` in that call shape, so method, headers (including auth) and body vanish and the call goes out as a bare GET. No current caller passes a `Request`, which is exactly why this will not be noticed when one does. Either narrow the exported type to `(url: string | URL, init?: RequestInit)` or read the `Request` properly.
* **`deleteConnector` swallows the credential delete.** `queries.deleteCredential(...).catch(() => {})` — a failure leaves behind precisely the orphan the module header says must never exist ("a credential appearing in the Setup list with no connector explaining it"), and with a live password in it, silently. At least log it; better, delete the credential first and let the connector delete be the compensable half, mirroring how `createConnector` orders its two writes.
* `createConnector` compensates a failed connector insert by deleting the credential, but that compensating delete is also `.catch(() => {})` — same gap, opposite direction.
* `verifyConnector` writes `configPatch` back into `config`. The patch is built from named fields rather than spread from the response, so the key set is safe; the `instanceUrl` *value* is attacker-influenced and becomes the target of later calls. `safeOutboundFetch` covers that, so this is a note rather than a finding — but a URL-shape check before persisting would make the guarantee local instead of remote.

mtolmacs (dismissed):

mtolmacs (approved):

PR #133, opened 2026-09-09
Add @lastest/veeva data migration package with full migration pipeline

ewyct:
## Summary

This PR introduces `@lastest/veeva-migration`, a comprehensive TypeScript package implementing a complete Salesforce-to-Veeva Vault data migration pipeline. The package handles extraction from SFDC, transformation, and bulk loading into Vault with extensive validation, error handling, and reconciliation capabilities.

## Key Changes

- **Core Migration Engine** (`src/run/engine.ts`, `src/cli.ts`): Multi-phase orchestration supporting preflight validation, extraction, transformation, and load operations with resumable checkpoints and comprehensive logging.

- **Preflight Validation** (`src/preflight/`): Offline and runtime checks including:
  - Configuration schema validation (`config/schema.ts`, `config/resolve.ts`)
  - Mapping lints and type compatibility matrix (`preflight/lints.ts`, `preflight/matrix.ts`)
  - Source/target connectivity and metadata validation
  - Legacy ID field resolution (`preflight/legacy-id.ts`)
  - Write probes for Vault permissions (`preflight/probes.ts`)
  - Crosswalk building for country/currency lookups (`preflight/crosswalks.ts`)

- **Extraction Pipeline** (`src/extract/`): 
  - Scope and country predicate evaluation (`extract/scope.ts`, `extract/country.ts`)
  - Delta window management with watermark tracking (`extract/delta.ts`)
  - SFDC REST/Bulk 2.0 query orchestration with PK chunking (`extract/extractor.ts`)
  - FK closure for referenced parent records (`extract/closure.ts`)
  - Client-side ordering for hierarchical loads (`extract/order.ts`)
  - CSV streaming with checkpoint recovery (`extract/files.ts`)

- **Object Modules** (`src/objects/`): 40+ Veeva CRM object definitions with:
  - Field mappings, transforms, and country attribution rules
  - Create/delete/match policies and blob handling
  - Block S system/audit row generation (`objects/types.ts`)
  - Topological load ordering with cycle detection (`objects/registry.ts`)
  - Reference objects: `user`, `territory`, `country` with specialized handling

- **Transform System** (`src/transform/`):
  - ID format conversion (15→18 char Salesforce IDs)
  - Custom transform functions (user status, country mapping, etc.)
  - Deferred reference resolution (`load/resolve-refs.ts`)
  - Type compatibility checking and auto-switching

- **Load Pipeline** (`src/load/`):
  - Bulk upsert with deduplication and hash-based skip detection (`load/loader.ts`)
  - Pre-existing record matching via VQL (`load/matcher.ts`)
  - Blob field deferred writes (`load/blobs.ts`)
  - Delete routing and inactivation policies (`load/deletes.ts`)
  - Second-pass self-reference patching (`load/second-pass.ts`)
  - Retry logic with exponential backoff (`load/retry.ts`)
  - Pending FK queue for unresolved references (`load/pending.ts`)

- **Configuration & State** (`src/config/`, `src/store/`):
  - YAML config parsing with environment variable interpolation
  - Multi-level config layering (global → region → country)
  - File-based and PostgreSQL state stores for run tracking
  - Mapping hash computation for change detection

- **API Clients** (`src/sfdc/`, `src/vault/`):
  - Salesforce: JWT auth, REST queries, Bulk 2.0 API, SOQL builder
  - Veeva Vault: OAuth2, VQL queries, bulk records API, metadata fetch, MDL generation
  - Comprehensive error handling and retry strategies

- **Testing Infrastructure** (`src/testkit/`):
  - Fake SFDC/Vault clients for unit testing
  - Fixture builders for describes, mappings, and metadata
  - Memory-backed state store for test isolation

- **Documentation**:
  - `docs/MIGRATION_SPEC.md`: Normative specification (2000+ lines)
  - `docs/CONTRACTS.md`: Builder's guide

https://claude.ai/code/session_01Ng9jVNLdaEMhK1SbszL2rL

## Why this PR touches both `core/` and `plugins/` (`core-and-plugin`)

This is the first plugin to declare the `jobs` capability, and the second review found that the queue as it stood could not safely run a long job: `cancelPluginJob` only touched `pending` rows, nothing ever aborted a handler already in flight, and `plugin_jobs` had no lease, so a worker that died held its job as `running` forever and its dedupe key locked the project. Those are queue properties, so the fix has to live in `core/jobs` (worker heartbeat + `reapExpired` hook) and `packages/db` (`plugin_jobs.heartbeat_at`, guarded `complete`/`failAttempt`), and the plugin side (`reconcileStaleRuns`, `finishMigrationRun`, the partial unique index) only makes sense on top of it. Landing the core half alone would ship a lease nothing exercises; landing the plugin half alone would ship a cancel that does not cancel. They are one bootstrap, verified together by `e2e/plugin-jobs-lease.integration.test.ts`.


ewyct:

| -46 | 0 | 53 | 0 | 32 |

[View Build](http://localhost:3000/verify/e1cfc613-8e64-4150-adee-16d06af85782)

---

mtolmacs (dismissed):

mtolmacs (changes requested):
# Review: Veeva CRM → Vault CRM migration (PR #133)

Focus: tenancy, permission handling, data consistency. Follow-up to the previous architecture review.

## Verdict

The placement and tenancy work from the previous review is done and holds up. Engine state is keyed by project in every primary key, the store goes through the plugin data handle, deletion hooks exist, the run directory is derived, the acting user is passed to the engine, and `pnpm arch` reports zero target-layout violations.

The remaining problems are all in the **run lifecycle**. Three of them contradict claims made in code comments and in `docs/architecture/veeva-migration-migration-result.md`, and they matter because this feature writes to a customer's live Vault.

## Blocking

### 1. Cancelling a running migration does not stop it

`plugins/veeva-migration/src/actions.ts:501` states that cancel aborts the handler's signal. It does not.

- `cancelPluginJob` (`src/lib/db/queries/plugin-jobs.ts:88`) only flips rows whose status is `pending`.
- The only thing that ever aborts the worker's `AbortController` is `perJobTimeoutMs` (`core/jobs/src/worker.ts:69`), which the runtime never sets. Nothing polls job status.

Result: the engine keeps extracting and upserting, the run row says `aborted`, and the one-run-per-project guard opens. When the engine finishes, `plugins/veeva-migration/src/jobs.ts:262` overwrites the operator's `aborted` with `succeeded`/`failed`. This is the defect the first review raised, now with a comment saying it is fixed.

**Fix:** give cancel a path to the in-flight controller (worker registry keyed by job id, or the handler polls `ctx.jobs.status` and aborts its own controller), let `cancelPluginJob` mark `running` rows, and make the final status update conditional on the row still being `running`.

### 2. The stale-run sweeper aborts healthy runs after 30 minutes

`failStaleMigrationRuns` (`plugins/veeva-migration/src/data/queries.ts:371`) compares `startedAt` against a 30-minute cutoff. `startedAt` is set at enqueue and nothing in the plugin ever heartbeats it.

- A multi-hour init is flagged "server restarted" 30 minutes in, the guard releases, and the row flips back to `succeeded` hours later.
- The worker is sequential, so a run from another project queued behind a long one sits past the cutoff, is swept to `aborted`, and is **then executed anyway**: `executeRun` (`jobs.ts:166`) sets `running` without checking the row's current status. An operator who saw "aborted" gets a live load they did not expect.

**Fix:** reconcile against the `plugin_jobs` row's status rather than wall clock (or heartbeat from the handler at unit boundaries), and have `executeRun` refuse any run whose row is not `queued`.

### 3. A worker death locks the project permanently

`plugin_jobs` has no lease or heartbeat, and `claimDuePluginJobs` (`src/lib/db/queries/plugin-jobs.ts:105`) claims only `pending` rows, so a job whose process died stays `running` forever. The doc's claim that core "accounts for a claimed job whose worker died" is not true.

Because the dedupe check matches `running` as well as `pending` (`plugin-jobs.ts:43`), every later `enqueueMigrationRun` for that project silently gets the dead job back, and `jobs.ts:98-123` creates a fresh run row pointing at it that nothing will ever execute. The sweeper aborts that row 30 minutes later. The project cannot run anything again without hand-editing `plugin_jobs`.

The same check-then-insert race in `enqueueMigrationRun` produces phantom run rows when two operators click at once.

**Fix:** a lease with reclaim in core (honouring `maxAttempts: 1` so a reclaimed run fails rather than re-executes), and a partial unique index on `(project_id) WHERE status IN ('queued','running')`. At minimum, detect that enqueue returned an existing job and refuse instead of creating the phantom row.

## Should fix before merge

### 4. Cross-tenant deletion of finding acknowledgements

`revokeFinding` (`actions.ts:544`) guards the project, but `revokeFindingAck` (`data/queries.ts:446`) deletes by `id` alone. Any member with migration access on any project can delete another team's acknowledgement given its id. Add the `projectId` predicate.

### 5. `advanced` overrides the connector-derived hosts

`config-builder.ts:268` merges `advanced` last over every top-level key except `countries`, and `updateMigrationSettings` accepts the whole config blob from the client. `advanced.target = { vaultDns: "…" }` replaces the host, and `withSecrets` then attaches the real decrypted Vault credential to it.

Not a privilege escalation (a `repos:settings` member can already re-point a connector while keeping its stored secret, `src/server/actions/connectors.ts:242`), but it breaks the invariant the design rests on: hosts come from connectors, never from a form. Allowlist the keys the merge may touch (`objects`, `performance`, `scope`, …) and exclude `source`, `target`, `version`, `waves`.

### 6. Run artifacts are never deleted

`runArtifactRoot` writes extract pages (customer HCP/account data) under `STORAGE_ROOT/migrations/<team>/<project>/<run>`. `deleteMigration`, `onRepoDeleted` and `onTeamDeleted` remove rows only. The plugin cannot touch the filesystem by design, so the host needs a `removeArtifacts(teamId, projectId)` beside `runArtifactRoot`, called from the delete action and the deletion hook.

### 7. Endpoints can be swapped after runs exist

`updateMigrationEndpoints` has no stage or run check. Watermarks are keyed by project only, so pointing a project at a fresh Vault after an init means the next delta starts from the old high-water mark and skips everything before it. That is the UAT-then-PROD bug again, inside one project. Refuse endpoint changes once any run exists, or key watermarks by `vault_dns` too.

## Minor

- `scripts/migrate.js` drops an empty destination table without `CASCADE`. When an earlier push already created the new tables with their internal FKs (the doc says this happened locally), dropping `veeva_migration_projects` fails on the dependent waves table.
- `summary.reportPath` is an absolute server path rendered to the browser (`ui/run-panels.tsx:461`) and is not downloadable. Serve the report through the app or drop the path.
- `src/lib/core/scheduler.ts` imports one plugin by name for `reconcileStaleRuns`, and the `pluginJobsProcessing` flag means the reconciler never runs while a long job is in flight.
- All three `/migrations` pages catch every error and render `MigrationLocked`, so a database outage reads as "not Early Adopter".
- `veeva_migration_audit_log.actor` is not anonymised on user deletion. Documented and defensible, but it is a GDPR trade-off that should be confirmed deliberately.

## Confirmed fixed from the previous review

- Feature moved to `plugins/veeva-migration`; engine moved to `libs/veeva-migration`; no `@/` imports in the plugin.
- `projectId` leads every engine primary key (asserted by `schema.test.ts`); store built over `ctx.data`, no own connection, no DDL on page render.
- Deletion hooks for team, repo and user; engine tables cascade from the project row.
- `run_dir` column removed; path derived by the host.
- Acting user passed as `RunContext.actor`; `DATABASE_URL` no longer in engine config.
- `exports` map on the engine; root `tsconfig` back to ES2017; demo video and Python renderer removed.

## Verification

- After `pnpm install --frozen-lockfile`: plugin unit tests 48/48, engine run + store tests 102/102, `pnpm --filter @lastest/plugin-veeva-migration typecheck` clean, `pnpm arch` 0 target-layout violations.
- `e2e/veeva-migration-store.integration.test.ts` could **not** be verified here: the local database has no `veeva_migration_*` tables (branch schema not pushed), so 37/38 fail with `42P01`. `pnpm db:push` was not run against the dev database.

ewyct:
Changes from the second review are in (9f0e2c53, c2acf291). Ready for another look.

**Blocking**
1. **Cancel stops a running run.** `cancelPluginJob` now flips `running` rows; the core worker heartbeats each job's lease every 30 s and aborts the handler's `AbortController` when the row reads cancelled. Every terminal write to the run row goes through `finishMigrationRun` (conditional on still in flight), so the operator's `aborted` is never overwritten by a late verdict.
2. **Sweeper no longer judges by wall clock.** `reconcileStaleRuns` follows `ctx.jobs.status` per in-flight run; only a `queued` row with no job id is timed out. `executeRun` refuses any run whose row is not `queued` (`claimQueuedMigrationRun`). The reconciler ticks under its own scheduler flag, so it keeps running during a long job.
3. **Dead worker no longer locks the project.** `plugin_jobs.heartbeat_at` lease; `reapExpiredPluginJobLeases` runs before each claim and goes through `failPluginJobAttempt`, so `maxAttempts: 1` settles as `failed` rather than re-executing. `complete`/`failAttempt` only touch `running` rows. Partial unique index `uq_veeva_migration_runs_one_active` on `(project_id) where status in ('queued','running')`; `enqueueMigrationRun` deletes its own row and refuses if the dedupe key handed back a job another run owns.

**Should fix**
4. `revokeFindingAck` scoped by `projectId`.
5. `advanced` merge allowlisted (`ADVANCED_OVERRIDABLE_KEYS`); `source`/`target`/`version`/`waves` never come from the form, and per-country `target` is stripped from overlays. Test added.
6. `VeevaMigrationHost.removeArtifacts(teamId, projectId?)`, called from `deleteMigration`, `onRepoDeleted`, `onTeamDeleted`.
7. `updateMigrationEndpoints` refuses once any run exists.

**Minor**
- `migrate.js` drops the empty destination with `CASCADE`.
- `summary.reportPath` is stored relative to the artifact root; the UI no longer renders an absolute server path.
- `/migrations` pages render `MigrationLocked` only for gate errors (`isMigrationGateError`) and rethrow everything else.
- Scheduler still names the plugin; commented as the case for a manifest tick hook. `audit_log.actor` retention left as the documented decision.
- Doc §4.1 rewritten to describe the lifecycle as it now is.

**Verification (local, real dev DB after `pnpm db:push`)**
- `tsc` clean for app and plugin; `pnpm arch` 0 target-layout violations; eslint clean.
- Unit: core/jobs 9/9 (3 new), plugin 49/49 (1 new), packages/db 25/25, src/lib/core 17/17.
- Integration: `veeva-migration-store` 38/38 (previously unrunnable, tables missing); new `e2e/plugin-jobs-lease.integration.test.ts` 5/5 covering lease refresh, cancel-while-running, guarded settles, reap under `maxAttempts: 1`, dedupe-key release.
- Partial index confirmed by hand to reject a second active run with the constraint name `enqueueMigrationRun` checks.
- Not exercised: an end-to-end run through the UI (needs live Salesforce + Vault connectors).

